PR #7, opened 2017-03-31
SensorList.java

akarshan96:
SensorList.py ported to SensorList.java


viveksb007:
@akarshan96 This has already been ported. So you can close this PR.
First Open an issue which tells what are you working on like in this case "add sensorlist" , and tell that you are working on it,so that no one else end up doing that.

viveksb007:
@wavicles this PR was not to be merged as its completely redundant. 
Have a look, its already there.
here https://github.com/fossasia/pslab-android/blob/master/app/src/main/java/com/viveksb007/pslab/communication/SensorList.java

Issue #25, opened 2017-04-05
Applications Fragment UI


akarshan96:
![screen shot 2017-04-05 at 11 48 27 pm](https://cloud.githubusercontent.com/assets/26146675/24720346/83adf436-1a5a-11e7-9c91-28b94d2959a8.png)

This is a screenshot of the proposed design for Applications Fragment.

viveksb007:
@akarshan96 great looks good to me.

mariobehling:
Excellent.

wavicles:
@akarshan96 Good work :+1:

PR #29, opened 2017-04-06
Updated Application Fragment

akarshan96:
UI as showed in #25


viveksb007:
@akarshan96 there might be a problem with UX in this implementation of yours. 
You created separate ImageButtons and TextViews. Now OnClickListeners is set on ImageButton. So clicking on TextView would result in nothing.
This problem magnifies in tablets/phones with big screens.

Could you use ListView with CustomAdapter to implement the same? This would eradicate the problem plus it would be easy to modify layout in future. 
And please give meaningful ids to widgets like "tv_oscilloscope" instead of "textView1".
Thanks.

akarshan96:
@viveksb007 I have tested the implementation on the tablet before making a pr and it worked fine. Yes, we can use listview for the same, I will make changes and make a pr. 
Thanks.

viveksb007:
@akarshan96 you didn't understand the problem.
I am saying that you stacked all ImageButtons and TextView on one another. But when you will set OnClickListener. You will set it on some widget right like ImageButton or TextView ( take note, listener is set on anyone and logically it would be imageButton ). So In tablet textView would also have appreciable height and when user would click on it, nothing would happen. When user would click on ImageButton, its okay. Its just a bad UX.
Whereas if its implemented using ListView and Custom adapter, listener would be set to complete ListItem.

Did you get the problem now? I wasn't talking about how it would look on tablet.

PR #37, opened 2017-04-09
Rebuilt Applications Fragment according to ListView approach

akarshan96:


viveksb007:
@akarshan96 nice work.

PR #38, opened 2017-04-09
reformatted and cleaned code

viveksb007:
- optimised listView ( doesn't affect much in this case as there are only 8 list items )
- reformatted and cleaned code


mariobehling:
Which issue is this related to? Let's start to follow best practices, so we are all on the same page. Thank you.

viveksb007:
Sorry, was just cleaning code after akarshan's PR.
Will create an issue for the same next time.

Issue #54, opened 2017-04-17
Porting SHT21.py


akarshan96:

akarshan96:
@viveksb007 I think you missed porting [this](https://github.com/fossasia/pslab-python/blob/development/PSL/Peripherals.py#L228) method in i2c.java

viveksb007:
@akarshan96 Thanks for the finding. Solved it in my PR https://github.com/fossasia/pslab-android/pull/55

Issue #53, opened 2017-04-17
Porting TSL2561.py 


akarshan96:
Porting the files from Sensor which are not dependent on sciencelab.py

PR #58, opened 2017-04-18
Updated DACChannel, Ported some Sensor files

### LINKED ISSUE #51: faulty polynomial implementation in DACChannel.java
In DACChannel.java coefficients for 0 & 1 are put [see here](https://github.com/fossasia/pslab-android/blob/master/app/src/main/java/org/fossasia/pslab/communication/peripherals/DACChannel.java#L29)
And in MCP4728 coefficient for 2 is queried [see here](https://github.com/fossasia/pslab-android/blob/master/app/src/main/java/org/fossasia/pslab/communication/peripherals/MCP4728.java#L63), as its put in a map, it would return null and null in mathematical operation would raise error.

akarshan96:
- Updated DACChannel.java to fix #51 
- Ported some Sensor files
-Added TSL2561.java #53 
-Added MLX90614.java #57 
-Added SHT21.java #54 
-Added BMP180.java #56

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/communication/sensors/BMP180.java
+++ b/app/src/main/java/org/fossasia/pslab/communication/sensors/BMP180.java
@@ -1,0 +1,151 @@
+package org.fossasia.pslab.communication.sensors;
+
+import android.util.Log;
+import org.fossasia.pslab.communication.peripherals.I2C;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.concurrent.TimeUnit;
+import static java.lang.Math.pow;
+
+/**
+ * Created by akarshan on 4/18/17.
+ */
+
+public class BMP180 {
+
+    private int ADDRESS = 0x77;
+    private int REG_CONTROL = 0xF4;
+    private int REG_RESULT = 0xF6;
+    private int CMD_TEMP = 0x2E;
+    private int CMD_P0 = 0x34;
+    private int CMD_P1 = 0x74;
+    private int CMD_P2 = 0xB4;
+    private int CMD_P3 = 0xF4;
+    private int oversampling = 0;
+
+    private int NUMPLOTS = 3;
+    String[] PLOTNAMES = {"Temperature", "Pressure", "Altitude"};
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/fossasia/pslab/communication/sensors/BMP180.java
```diff
@@ -0,0 +1,151 @@
...
+        params.put("setOversampling",new ArrayList(Arrays.asList(0,1,2,3)));
+        initTemperature();
+        readTemperature();
+        initPressure();
+        baseline = readPressure();
+    }
+
+    int readInt(int addr)
+    {
+        return(Integer.parseInt(Integer.toString(addr).trim()));
```
viveksb007: what are you trying to do here?
From what i understand, you are taking an integer, converting it to string and trimming it ( why? ) and again parsing it to string.
In python library, https://github.com/fossasia/pslab-python/blob/development/PSL/SENSORS/BMP180.py#L57 
Its reading an integer at address provided and returning it.

app/src/main/java/org/fossasia/pslab/communication/sensors/BMP180.java
```diff
@@ -0,0 +1,151 @@
...
+            return null;
+        }
+    }
+
+    double altitude()
+    {
+        // baseline pressure needs to be provided
+        return (44330.0 * (1 - pow( PR / baseline, 1 / 5.255)));
+    }
+
```
viveksb007: Coding style doesn't match the style followed in entire repo, instead of (double P, double A) you can write (double pressure,double altitude) or something like that helps in understanding the code and concept better. ( If something isn't clear, put a comment )

app/src/main/java/org/fossasia/pslab/communication/sensors/BMP180.java
akarshan96: Here I am converting hexadecimal to decimal.
`System.out.println(Integer.parseInt(Integer.toString(0xBA)));`
Output: 186

akarshan96: Got that wrong thanks for pointing.

PR #66, opened 2017-04-21
Created SSD1306, Reformatted code

akarshan96:
- Created SSD1306 and added some change #61 
- Reformatted and cleaned some code
- Made few suggested changes


app/src/main/java/org/fossasia/pslab/communication/sensors/SSD1306.java
```diff
@@ -0,0 +1,469 @@
...
+
+public class SSD1306 {
+    public int ADDRESS = 0x3C;
+    public HashMap<String, ArrayList> params = new HashMap<>();
+    public int NUMPLOTS = 0;
+    public String[] PLOTNAMES = {""};
+    public String name = "OLED Display";
+
+    private int _width = 128;
+    private int WIDTH = 128;
```
viveksb007: remove "_width" and "_height"

akarshan96: I left that to avoid confusion with the WIDTH. However, they are having different case but it felt that it should remain _width.

viveksb007: What confusion with WIDTH, I didn't even see any use of WIDTH, you were supposed to replace '_width' with either 'width' or 'WIDTH', same with '_height' . 
either remove '_' from '_width' or use WIDTH.
Its doesn't matter much, just to have uniformity in code.

akarshan96:
Suggested changes are made :)

PR #70, opened 2017-04-27
Ported some sensor files

### LINKED ISSUE #64: port MPU6050.py

### LINKED ISSUE #65: Port MF522.py

akarshan96:
- Created MPU6050.java, fixes #65 
- Created MF522.java, fixes #64 
- Added Filters (Kalman Filter and Complementary Filter)
- Added Apache Commons Maths Library for scientific calculations


app/src/main/java/org/fossasia/pslab/communication/sensors/MPU6050.java
```diff
@@ -0,0 +1,136 @@
...
+                }
+                return raw;
+            }
+        }
+        return null;
+    }
+
+    public double[] getAcceleration() throws IOException {
+        ArrayList<Character> vals = getVals(0x3B,6);
+        int ax = int16(vals.get(0) << 8 | vals.get(1));
```
viveksb007: I am not sure if this does anything finally.
you are giving and "int" to int16 to convert it to short and then again putting it into "int". So number of bits are same as before.
remove short, we would work with "int" everywhere. As it just a matter of number of bits and Lower bit order is same in both short and int.

P.S: when implemented something just do a dry run to get to know what its doing.

akarshan96: @viveksb007 the suggested are made

akarshan96:
@viveksb007 the suggested are made

akarshan96:
@mariobehling thanks for the update :)

PR #82, opened 2017-05-04
Fixes #75 ports ADS1115.py

### LINKED ISSUE #75: Port ADS1115.py

asitava1998:
Please suggest changes & let me know errors/improvements

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/communication/sensors/ADS1115.java
+++ b/app/src/main/java/org/fossasia/pslab/communication/sensors/ADS1115.java
@@ -1,0 +1,238 @@
+/**
+ * Created by asitava1998 on 3/5/17.
+ */
+
+package org.fossasia.pslab.communication.sensors;
+import org.fossasia.pslab.communication.peripherals.I2C;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.concurrent.TimeUnit;
+
+public class ADS1115{
+    private int ADDRESS = 0x48;
+    private I2C i2c;
+
+    private int REG_POINTER_MASK = 0x3;
+    private int REG_POINTER_CONVERT = 0;
+    private int REG_POINTER_CONFIG = 1;
+    private int REG_POINTER_LOWTHRESH = 2;
+    private int REG_POINTER_HITHRESH = 3;
+
+    private int REG_CONFIG_OS_MASK = 0x8000;
+    private int REG_CONFIG_OS_SINGLE = 0x8000;
+    private int REG_CONFIG_OS_BUSY = 0x0000;
+    private int REG_CONFIG_OS_NOTBUSY = 0x8000;
+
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/fossasia/pslab/communication/sensors/ADS1115.java
```diff
@@ -0,0 +1,243 @@
...
+
+        if(chan.equals("01"))
+            config = config | REG_CONFIG_MUX_DIFF_0_1;
+        else if (chan.equals("23"))
+            config = config | REG_CONFIG_MUX_DIFF_2_3;
+
+        //Set 'start single-conversion' bit
+        config = config | REG_CONFIG_OS_SINGLE;
+        writeRegister(REG_POINTER_CONFIG, config);
+        TimeUnit.SECONDS.sleep((long) (1./rate + 0.002));       //convert to mS to S
```
akarshan96: If you parse 1. / rate + 0.002 to long, it will always return 0 taking all possible rate values into consideration as long is an integer datatype.
Use the following approach.
Use TimeUnit.MILLISECOND.sleep((long)((1. / rate + 0.002) * 1000))
If I take rate as 8, the sleep time will be 0.127 seconds or 127 milliseconds.
Change the code using the similar approach in line 200 and 147.

app/src/main/java/org/fossasia/pslab/communication/sensors/ADS1115.java
```diff
@@ -0,0 +1,243 @@
...
+        sdrSelection.put(32,REG_CONFIG_DR_32SPS);
+        sdrSelection.put(64,REG_CONFIG_DR_64SPS);
+        sdrSelection.put(128,REG_CONFIG_DR_128SPS);
+        sdrSelection.put(250,REG_CONFIG_DR_250SPS);
+        sdrSelection.put(475,REG_CONFIG_DR_475SPS);
+        sdrSelection.put(860,REG_CONFIG_DR_860SPS);
+
+    }
+
+    public short __readInt__(int addr) throws IOException, InterruptedException {
```
akarshan96: Avoid the use of underscores for naming functions and variables in Java. Make such changes throughout the code.

viveksb007: one more thing readUInt is returning a "short" which readInt is casting as a "short" ( doesn't make any difference ) and returning again.
change readUInt to readInt ( get rid of readInt function ) and return normal "int" as lower bits of short are still there in int and we have been using "int" for porting.
This function is similar to readRegister in which you read 2 bytes and returned "int" from them.  

Thanks

asitava1998:
Fixed the changes & squashed the commits

viveksb007:
@asitava1998  squash again, my commit is also coming in your PR.
Save a local  copy of repo, in case something goes wrong.
Thanks

PR #84, opened 2017-05-05
Ported some sensor files

### LINKED ISSUE #77: Port HMC5883L.py

### LINKED ISSUE #78: Port AD9833.py

akarshan96:
- Created AD9833.java fixes #78 
- Created HMC5883L.java fixes #77

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/communication/sensors/AD9833.java
+++ b/app/src/main/java/org/fossasia/pslab/communication/sensors/AD9833.java
@@ -1,0 +1,108 @@
+package org.fossasia.pslab.communication.sensors;
+
+
+import android.util.Log;
+
+import org.fossasia.pslab.communication.peripherals.SPI;
+import java.io.IOException;
+
+
+/**
+ * Created by akarshan on 4/23/17.
+ * <p>
+ *     ScienceLab instance of SPI and DDS_CLOCK are required to be passed to the AD9833 constructor.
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/pslab/communication/sensors/AD9833.java
```diff
@@ -0,0 +1,106 @@
...
+    private int DDS_RESERVED = (1 << DDS_OPBITEN) | (1 << DDS_MODE);
+    private int  clockScaler = 4;       // 8MHz
+    private int waveformMode;
+    private int activeChannel;
+    private int frequency;
+    private int cs;
+    private ScienceLab scienceLab;
+    
+    public AD9833(UsbManager usbManager) throws IOException{        //this approach needs to be checked.
+        scienceLab = new ScienceLab(usbManager);
```
viveksb007: Don't create a new scienelab object, if needed I would pass sciencelab object from activity ( or wherever I will make object of this class ).
What i am thinking is I would pass sciencelab.spi to AD983's constructor and I suppose mapReferenceClock is only sciencelab method used, so would call it before making AD983's object.
Works?

akarshan96: Yes, this works. One more thing, mapReferenceClock would change the value of DDS_CLOCK. 
you would also need to pass the new value of DDS_CLOCK in the AD9833's constructor as it is used in AD9833.java.

akarshan96:
Requested changes are made !

PR #111, opened 2017-05-10
Fixes #110 Added boilerplate code for some activities

### LINKED ISSUE #110: Add boilerplate code for activities
Add boilerplate code for activities that are exposed by Application grid and obtain a reference of scienceLab instance inside those Activities.

viveksb007:
Added classes for below Applications 

- Oscilloscope
- Control
- Advanced Control
- Logical Analyzer

Would add more when these would be implemented to some extent.

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/fragment/ApplicationsFragment.java
+++ b/app/src/main/java/org/fossasia/pslab/fragment/ApplicationsFragment.java
@@ -1,6 +1,7 @@
 package org.fossasia.pslab.fragment;
 
 import android.content.Context;
+import android.content.Intent;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/fossasia/pslab/fragment/ApplicationsFragment.java
```diff
@@ -40,18 +45,36 @@ public static ApplicationsFragment newInstance() {
...
-            @Override
-            public void onItemClick(ApplicationItem item) {
-                Toast.makeText(context, item.getApplicationName(),
-                        Toast.LENGTH_LONG).show();
-            }
-        });
+                    @Override
+                    public void onItemClick(ApplicationItem item) {
+                        Toast.makeText(context, item.getApplicationName(),
+                                Toast.LENGTH_LONG).show();
```
CloudyPadmal: These lines can be removed now right?  :)

viveksb007: Yeah, but I kept them as I haven't added classes for some applications like sensor  quick view, data sensor logger, etc. So kept it as an indication that they have been clicked.

CloudyPadmal: Then it's okay!! :D

PR #124, opened 2017-05-21
Fixes #116 implements butterknife

### LINKED ISSUE #116: Add butterknife for view injections
Add [Butterknife](https://github.com/JakeWharton/butterknife) for view injections.

asitava1998:
Although in some activity files, views have not been defined yet, Butterknife has been added as views would be added in future definitely. In these files there is an import statement and a statement Butterknife.bind(this) which would work when views are defined for these files/

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -44,6 +44,8 @@ dependencies {
     compile 'com.android.support:recyclerview-v7:25.2.0'
     // Picasso
     compile 'com.squareup.picasso:picasso:2.5.2'
+    compile 'com.jakewharton:butterknife:8.6.0'
+    annotationProcessor 'com.jakewharton:butterknife-compiler:8.6.0'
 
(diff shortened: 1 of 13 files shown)


build.gradle
```diff
@@ -3,9 +3,11 @@
 buildscript {
     repositories {
         jcenter()
+        mavenCentral()
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:2.2.3'
+        classpath 'com.jakewharton:butterknife-gradle-plugin:8.6.0'
```
CloudyPadmal: Is this file necessary?

asitava1998: In the documentation it was included. Is it running without this file? Is the Travis check passing? If yes, then not necessary.

viveksb007: Its not necessary, works fine with adding in app level gradle file.

PR #132, opened 2017-05-30
Fixes #104  calibration method completed

### LINKED ISSUE #104: Complete calibration part in runInit function of sciencelab
Complete runInit function of science lab to successfully read and process calibration data.
https://github.com/fossasia/pslab-android/blob/master/app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java#L108

After this we can check other functionality of other applications by logging and processing data applying calibration.

viveksb007:
Fixes #104 Calibration method completed.
Calibration method is ported completely.

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
+++ b/app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
@@ -333,7 +333,30 @@ public class ScienceLab {
                             }
                         }
 
-                        // todo : port remaining function
+                        ArrayUtils.reverse(fitVals);
+                        PolynomialFunction fitFunction = new PolynomialFunction(fitVals);
+                        double[] yData = new double[DACX.length];
+                        for (int j = 0; j < DACX.length; j++) {
+                            yData[j] = fitFunction.value(DACX[j]) - (OFF.get(j) * slope + intercept);
+                        }
+                        int LOOKBEHIND = 100;
+                        int LOOKAHEAD = 100;
+                        ArrayList<Double> offset = new ArrayList<>();
+                        for (int j = 0; j < 4096; j++) {
+                            double[] temp = new double[Math.min(4095, j + LOOKAHEAD) - Math.max(j - LOOKBEHIND, 0)];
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
```diff
@@ -333,7 +333,30 @@ private void runInitSequence(Boolean loadCalibrationData) throws IOException {
...
 
-                        // todo : port remaining function
+                        ArrayUtils.reverse(fitVals);
+                        PolynomialFunction fitFunction = new PolynomialFunction(fitVals);
+                        double[] yData = new double[DACX.length];
+                        for (int j = 0; j < DACX.length; j++) {
+                            yData[j] = fitFunction.value(DACX[j]) - (OFF.get(j) * slope + intercept);
+                        }
+                        int LOOKBEHIND = 100;
+                        int LOOKAHEAD = 100;
```
CloudyPadmal: Hi @viveksb007 ! If both these values are having the same value, can we assign a single variable instead of two and use it in the code?

viveksb007: Yeah, it might have some significance in terms of calibrating the raw values. Yes we can assign a common variable and use it, currently we are taking 100 values from both ahead and back of current value. That might have yielded best results. Rest @jithinbp can clear as it was like this in python-communication-library.

jithinbp: both values can be assigned a single variable. But i'm not sure that the algorithm is the most efficient.

The basic goal of the algorithm is to locate the DAC value that will yield a voltage closest to the requested one. 
dac error
As you can see, digital to analog convertors deviate from the theoretical setpoint in an irregular manner. The PSLab stores the error present in each DAC code in an array of 4K elements (12-bit = 4096 codes) .
The object of the algorithm is to search for codes in the vicinity of the requested code in order to yield an output that is closest to the requested setpoint.   And as is evident from the yellow plot, this approach works fine and manages to get full 12-bit performance from the DAC. 

But because the calibration process is so long, we should probably switch to a quadratic polynomial correction for each DAC channel. The maximum error one can expect after quadratic correction is not more than 5mV .

Issue #164, opened 2017-06-13
cache gradle directory in travis


viveksb007:
cache gradle directory in travis to speed up the build process.

CloudyPadmal:
Currently it's downloading all the files from the beginning! This issue was addressed in fossasia-phimpme once! :)

viveksb007:
Yes, open-event-android also caches gradle directories. We can adopt caching code for travis.yml directly from there.

CloudyPadmal:
Let's enhance our one as well! :+1:

viveksb007:
Fixed in #171

PR #171, opened 2017-06-14
modified tavis.yml for enable cache

viveksb007:
Fixes issue #164 

Changes: Modified travis.yml file for cache.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,16 @@ licenses:
     - 'google-gdk-license-.+'
     - '.+'
 
+before_cache:
+  - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
+
+cache:
+  directories:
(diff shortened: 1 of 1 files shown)


viveksb007:
I think phimpme guys also copied their cache code from open-event-android. https://github.com/fossasia/phimpme-android/blob/development/.travis.yml#L56  because they don't have same project structure as open-event but directories path is same in travis. Like open-event-android have **app** folder inside **android** folder, whereas phimpme project doesn't.
@CloudyPadmal could you confirm the finding?

asitava1998:
How did you check if the modified travis.yml file works or not ?

CloudyPadmal:
It seems so @viveksb007 ! Project structure looks different in both repos. I went through this article regarding [Travis caching](https://docs.travis-ci.com/user/caching/#Arbitrary-directories) and the structure for our repo seems legit in your fix! :+1:

Issue #169, opened 2017-06-14
Show Channel Name as label on Y axis in Logic Analyzer


viveksb007:
**Actual Behaviour**

Currently there are numbers/values showing as labels on Y - axis of chart. 

**Expected Behaviour**

There should be channel name from which change in Logic is represented by Logic Line.

**Steps to reproduce it**

Open Logic Analyser and Click Analyze.

**LogCat for the issue**

None

**Screenshots of the issue**

![la](https://user-images.githubusercontent.com/12713808/27139739-6da11b64-5141-11e7-9434-cc133299b3aa.png)

**Would you like to work on the issue?**

Yes

viveksb007:
Fixed in #177

Issue #173, opened 2017-06-15
Modify Logic Analyzer UI parameters to match pslab-desktop-app


viveksb007:
**Actual Behaviour**

Currently there are options available for

- all channels in four channel mode
- Only four channels are available for single channel mode

**Expected Behaviour**

Match all such parameters to match logic analyser functionality of pslab-desktop-app.

**Steps to reproduce it**

Open Logic Analyser Activity to inspect.

**LogCat for the issue**

None 

**Would you like to work on the issue?**

Yes

viveksb007:
Fixed in #177

PR #177, opened 2017-06-15
Logic Analyzer UI modified as per pslab-desktop-app and labels indicate channel

viveksb007:
Fixes issue #169  #173 

Changes: 

- Modified Channel selection at different modes as per pslab-desktop-app
- Named logic lines according to which Channel was selected in Channel Mode select Fragment

Screenshots for the change: 
For Three channel mode, input channels are locked to ID1, ID2, ID3
screenshot_2017-06-15-21-54-45-167_org fossasia pslab

In two Channel Mode, **SEN** & **CNTR** are selected which are passed to next fragment.
screenshot_2017-06-15-21-55-00-978_org fossasia pslab
screenshot_2017-06-15-21-55-05-710_org fossasia pslab

Added time label also in latest commit.
screenshot_2017-06-15-23-03-25-484_org fossasia pslab

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
+++ b/app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
@@ -19,7 +19,7 @@ import com.github.mikephil.charting.data.LineDataSet;
 import com.github.mikephil.charting.interfaces.datasets.ILineDataSet;
 
 import org.fossasia.pslab.R;
-import org.fossasia.pslab.items.ChannelAxisFormatter;
+import org.fossasia.pslab.others.ChannelAxisFormatter;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -36,6 +36,7 @@ public class LALogicLinesFragment extends Fragment {
     private int channelMode;
     private Context context;
     private LineChart logicLinesChart;
+    private ArrayList<String> channelNames = new ArrayList<>();
 
     public static LALogicLinesFragment newInstance(Bundle params, Context context) {
         LALogicLinesFragment laLogicLinesFragment = new LALogicLinesFragment();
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
```diff
@@ -66,6 +90,18 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
         super.onViewCreated(view, savedInstanceState);
         updateLogicLines();
+        YAxis left = logicLinesChart.getAxisLeft();
+        left.setValueFormatter(new ChannelAxisFormatter(channelNames, channelMode));
+        left.setGranularity(1f);
+        left.setTextColor(Color.BLACK);
+        left.setTextSize(12f);
+        logicLinesChart.getAxisRight().setDrawLabels(false);
+
+        /*  For HIDING GRID LINES
```
CloudyPadmal: Do we need this commented lines?

app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
```diff
@@ -46,6 +49,27 @@ public static LALogicLinesFragment newInstance(Bundle params, Context context) {
...
+                channelNames.add(params.getString("inputChannel2"));
+                channelNames.add(params.getString("inputChannel3"));
+                break;
+            case 4:
+                channelNames.add(params.getString("inputChannel1"));
+                channelNames.add(params.getString("inputChannel2"));
+                channelNames.add(params.getString("inputChannel3"));
+                channelNames.add(params.getString("inputChannel4"));
+                break;
+            default:
```
CloudyPadmal: This can be simplified with removing break and arranging this in reverse order! :) Have a look at this [link](https://stackoverflow.com/questions/8563970/switch-without-break)

app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
viveksb007: Yes, when I am thinking of adding more settings like hiding grid lines,etc. So that's for it. I

app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
viveksb007: Yeah, its simplified but that doesn't solve the purpose. I need to add channels in order to populate in the graph in same order.
I can use your way and reverse the list, but that would make code unnecessary difficult to understand for some new dev + reversing the list has its overhead too.

CloudyPadmal: Yes @viveksb007 that's a good point! It will be confusing! :thinking:

app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
```diff
@@ -59,13 +86,27 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
         logicLinesChart.setDrawBorders(true);
         logicLinesChart.setBorderWidth(2);
         llLogicLines.addView(logicLinesChart);
+        tvTimeUnit = (TextView) v.findViewById(R.id.la_tv_time_unit);
+        tvTimeUnit.setText("Time -> (mS)");
```
CloudyPadmal: Can we make `mS` as `ms`? :)

wavicles: ms is correct way :)

viveksb007: yea, didn't notice that. Thanks

Issue #178, opened 2017-06-15
Main Control window having screen compatibility issues


CloudyPadmal:
**Actual Behaviour**

There is a text label at the right side of the `control_main_list_item.xml` but it is not provided with enough space on a `480 x 800` display

**Expected Behaviour**

The screen space should be allocated properly to display all the text fields

**Steps to reproduce it**
-    Open app
-   Go to `Applications`
-   Click on `Control` and go to main control section and observe

**LogCat for the issue**

> Not available

**Screenshots of the issue**

![Image](https://user-images.githubusercontent.com/14261304/27192011-375a6f6c-51e9-11e7-8b38-938b06a39ec8.png)

**Would you like to work on the issue?**

@asitava1998 will work on screen compatibility :+1:

viveksb007:
Fixed in #181

PR #181, opened 2017-06-16
Fix #176 : Main Control Panel UI Fixes

### LINKED ISSUE #176: control_main_list_item has a few issues in the view
**Actual Behaviour**

The `control_main_list_item.xml` ;
- Doesn't contain padding on left and right
- Text boxes allow letters other than numerical values
- Text boxes don't have a padding

**Expected Behaviour**

- Cardview needs a padding on the sides
- Text boxes can be allowed to have numerical values to avoid programming errors occurring in future when functionalities are to be implemented
- Text boxes should have a padding to display content not sticked to the border

**Steps to reproduce it**

- Open app
- Go to `Applications`
- Click on `Control` and go to main control section and observe

**LogCat for the issue**

> Not available

**Screenshots of the issue**

a

Yes!

CloudyPadmal:
Fixes issue #176 

Changes:
- Added padding to the list item
- Set type as decimals in text boxes
- Exported dimensions to `dimens.xml` so that it is easy to set dimensions for different screen sizes at one place
- Added padding to text boxes

Screenshots for the change: 
fix

### DIFF
--- a/app/src/main/res/layout/control_main_list_item.xml
+++ b/app/src/main/res/layout/control_main_list_item.xml
@@ -5,14 +5,16 @@
     android:layout_height="wrap_content"
     android:orientation="vertical"
     android:visibility="visible"
-    android:layout_marginTop="10dp">
+    android:layout_marginTop="@dimen/control_card_margin"
+    android:layout_marginEnd="@dimen/control_card_margin"
+    android:layout_marginStart="@dimen/control_card_margin">
 
     <android.support.v7.widget.CardView
         android:id="@+id/cardview_control_main"
         android:layout_width="match_parent"
-        android:layout_height="110dp"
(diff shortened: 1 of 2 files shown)


viveksb007:
Doesn't the seekbar looks a bit odd in the screenshot. The dot looks separated from the line.
In my phone its okay. What could be the reason for yours?

CloudyPadmal:
Yes @viveksb007 it is not having enough space so that it's adjusted a little upwards. @asitava1998 will you address it in #178 because it is a screen compatibility issue or should I address it here? :)

asitava1998:
Yes. i will address this in my PR.  So far I had only tested it for resolutions of 1920x1080, 1280x720 and 2048x1536. So, I wasn't aware of this. I will look into it.

PR #188, opened 2017-06-17
Fixes #165 Creates a simple starter UI for Sensors

### LINKED ISSUE #165: Create a generic application for reading data from I2C sensors
Field tests are necessary to test if the communication libraries for the sensors are working. 
it's easy to miss things while porting code, so  I propose to add a minimal UI which can auto-scan the I2C bus, let the user select a sensor, and fetch the data returned by the getRaw function.

A simple UI to quickly verify all the ported libraries should have these essentials.
+ autoscan (button) onClick -> display detected addresses
+ address (hexadecimal input / combobox with results of autoscan)
+ init (button) onClick -> Create an instance of the class associated with the selected address.
+ getRaw (button) onClick -> display array of data returned by the getRaw function

If we have the resources, we can make it slightly more sophisticated by adding the sensor-specific-functions in the form of widgets. But this might take a while, and is not critical for a minimal testing tool.
+ read the params HashMap, and create corresponding widgets
+ Associate callbacks to the corresponding functions in the sensor's class.

asitava1998:
Fixes issue #165 

Changes: 
* Creates a simple UI for Sensors.
* There is an auto scan button to detect sensors, the sensors detected will be displayed in the text view below & listed in the spinner.
* The sensor is selected from the spinner and then the get raw button is pressed to fetch raw data from the sensors.
* Detailed description of purpose and features is presented in issue description.

Screenshots for the change:


app/src/main/java/org/fossasia/pslab/activity/ReportUs.java
```diff
@@ -17,7 +17,7 @@
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        setContentView(R.layout.activity_report_us);
```
viveksb007: Please change this back.
Probably a mistake

app/src/main/res/values/strings.xml
```diff
@@ -84,4 +84,10 @@
     <string name="digital_outputs">Digital Outputs</string>
     <string name="configure_PWM">Configure PWM</string>
     <string name="waveform_generator">Arbitrary Waveform Generator</string>
+    <string name="autoscan">AUTOSCAN</string>
+    <string name="select_sensor">SELECT SENSOR</string>
+    <string name="get_raw">GET RAW</string>
+    <string-array name="sensors">
+
```
viveksb007: sensor would be scanned on the fly. So why store an empty string array here?

app/src/main/res/layout/activity_sensor.xml
```diff
@@ -0,0 +1,71 @@
...
+            android:layout_height="wrap_content"
+            android:layout_marginTop="20dp">
+            <TextView
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:text=""
+                android:background="@drawable/control_custom_border"/>
+        </ScrollView>
+    </LinearLayout>
+</ScrollView>
```
CloudyPadmal: Hello @asitava1998 this fix is really nice! When you're changing the things Vivek has pointed out, can you add a new line here? :blush:

app/src/main/res/values/strings.xml
```diff
@@ -84,4 +84,10 @@
...
+    </string-array>
 </resources>
```
CloudyPadmal: and here as well!

app/src/main/res/values/strings.xml
```diff
@@ -84,4 +84,10 @@
     <string name="digital_outputs">Digital Outputs</string>
     <string name="configure_PWM">Configure PWM</string>
     <string name="waveform_generator">Arbitrary Waveform Generator</string>
+    <string name="autoscan">AUTOSCAN</string>
+    <string name="select_sensor">SELECT SENSOR</string>
+    <string name="get_raw">GET RAW</string>
```
CloudyPadmal: We can undo these strings to be in normal case and set attributes in buttons to view them as uppercase! What do you think? :)

app/src/main/res/values/strings.xml
asitava1998: There is a layout rendering error if I don't reference a string array in the spinner.

app/src/main/res/layout/activity_sensor.xml
asitava1998: Okay. sure

app/src/main/res/values/strings.xml
asitava1998: :+1:

app/src/main/res/values/strings.xml
asitava1998: :+1:

app/src/main/java/org/fossasia/pslab/activity/ReportUs.java
asitava1998: Yes. might have done some copy pasting in the wrong file accidently. I will correct it.

app/src/main/java/org/fossasia/pslab/activity/SensorActivity.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+public class SensorActivity extends AppCompatActivity {
+    private ScienceLab scienceLab;
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_sensor);
+        scienceLab = ScienceLabCommon.getInstance().scienceLab;
```
viveksb007: now that sciencelab is static you don't need to get instance.
```sciencelab = ScienceLabCommon.scienceLab``` would do the job.

asitava1998: I will correct in the next PR regarding the sensor UI where functions will be added

PR #189, opened 2017-06-17
Fixes #148 creates UI for Control Main

### LINKED ISSUE #148: Control UI Implementation - Basic Control
The parent discussion can be found here #41 . The parent discussion has some valuable discussions on points about the UI design. This thread contains functionalities for using PSLab as a voltage/current source and as a wave generator.

-  [ ] Layout design
-  [ ] Code for adding functionalities

The mockup for the same is 
control2

The actual UI for the same is

There might be changes in the actual UI during the PR.

asitava1998:
Fixes issue #148 

Changes: 
* Added function to buttons except for set button which will be done when scienceLab functionalities will be added.
* Set the edittext to only consider decimals, set limits on data input according to range of PV1, PV2, PCS , W1 etc.
* Set the limits on seekbar as per range and synchronise the edittext as per the seekbar.
* Minor change in UI of control read.

Changes remaining:
* Adding remaining function to buttons.
* Synchronise the seekbar as per the edittext( vice-versa of last change).

Screenshots for the change:


viveksb007:
@asitava1998 what about making bottom tab-strip, indicative of which fragment is selected?

app/src/main/java/org/fossasia/pslab/adapters/ControlMainAdapter.java
```diff
@@ -13,17 +17,37 @@
  * Created by asitava on 6/6/17.
  */
 
-public class ControlMainAdapter extends RecyclerView.Adapter<ControlMainAdapter.ViewHolder> {
+public class ControlMainAdapter extends RecyclerView.Adapter<ControlMainAdapter.ViewHolder> implements View.OnClickListener{
```
viveksb007: If you are setting ```OnClickListener``` individually to all view elements in ```onBindViewHolder```, then why implement ```OnClickListener``` to complete adapter.

app/src/main/res/layout/control_main_list_item.xml
```diff
@@ -20,63 +20,66 @@
...
+                android:layout_width="0dp"
+                android:layout_weight="3"
+                android:layout_height="30dp" />
         </LinearLayout>
 
         <SeekBar
-            android:id="@+id/seekbar_control_main1"
+            android:id="@+id/seekbar_control_main"
             android:layout_width="match_parent"
             android:layout_height="50dp"
```
CloudyPadmal: What if we set this to `wrap_content`? That fixed the bubble rising above the line issue I had in my 480x800 screen :)

app/src/main/java/org/fossasia/pslab/adapters/ControlMainAdapter.java
asitava1998: The problem I was facing is I have three buttons with ```OnClickListener``` and only one ```OnClickListener``` for the buttons is actually working, whereas in case of edittext and seekbar, since are just one in number they work fine. So, i wrote it.

app/src/main/res/layout/control_main_list_item.xml
asitava1998: Okay sure. This would also fix another issue.
:+1:

asitava1998:
@viveksb007  I was thinking of making the bottom navigation bar indicative. But it would lead to this PR solving 3-4 issues at the same time. If this is not a problem, I can make the change. Let me know

viveksb007:
Okay then, solve it in another PR :+1:

app/src/main/java/org/fossasia/pslab/adapters/ControlMainAdapter.java
viveksb007: removing it won't make a difference because you are not doing anything in ```onClick``` Line 49. So try removing this ```implements View.OnClickListener``` shouldn't make a difference.

Issue #194, opened 2017-06-18
scienceLab object is re-initialised when orientation is changed


viveksb007:
**Actual Behaviour**

scienceLab object is re-initialised when orientation is changed. ( portrait -> landscape & vice-versa )

**Expected Behaviour**

scienceLab object should not be re-initialised when orientation is changed

**Steps to reproduce it**

Go to MainActivity ( HomeFragment ) and change orientation and see logs or it would reflect on screen too. 

**LogCat for the issue**

None

**Would you like to work on the issue?**

Yes

viveksb007:
Fixed in #198

PR #198, opened 2017-06-19
Fixes re-initialisation on screen rotation

viveksb007:
Fixes issue #194 

Changes: State of Activity is retained on screen rotation so that scienceLab object doesn't re-initialises itself.


mariobehling:
Good to have this working now!

Issue #72, opened 2017-04-27
Handling abrupt disconnection/device not found cases


asitava1998:
Addition of a manual Retry/Reconnect button in the Home UI if connection fails even when the device is connected.
While in operation, if the device gets abruptly disconnected, two solutions can be done-
1. Landing back on Home UI and trying to reconnect again.
2. Displaying a toast asking user to manually reconnect again.

Although its a minor issue, I would like to do it to get started and also get a hang of this project.

viveksb007:
@asitava1998 Great, start working on it and send a PR.
I was thinking, we should have a reconnect icon on the actionBar when UI is HomeFragment only or refresh by pulling HomeFragment down.

viveksb007:
@asitava1998 has there been any improvement for solving this issue?

asitava1998:
@viveksb007  I had added a button to refresh the connection in this issue. However, after that changes were made to how the sciencelab class works (provides instances), I still have the code with me, can make a few changes in it and send a PR.
However, swipe to refresh cannot be implemented in the home fragment as it is limited to ListView and RecyclerView and the home fragment does not use these views.

viveksb007:
@asitava1998 saw your scrum, I already worked on it yesterday. And after testing would send a PR today.

viveksb007:
Fixed in #217

Issue #216, opened 2017-06-22
Main thread and Initialisation thread routines are coinciding resulting in wrong version string


viveksb007:
**Actual Behaviour**

Main Thread (UI thread) and initialisation threads routines coincide when version read is done in between initialisation is going on and version string is set up as some part of calibration data.

**Expected Behaviour**

It shouldn't do a version read between initialisation.

**Steps to reproduce it**

Open App go to Application Fragment and switch back before initialisation is completed.

**LogCat for the issue**

None

**Screenshots of the issue**

![screenshot_2017-06-22-20-32-44-490_org fossasia pslab](https://user-images.githubusercontent.com/12713808/27441675-e6879aa8-578b-11e7-9bf8-754e98c92055.png)

**Would you like to work on the issue?**

Yes

viveksb007:
Fixed in #217

jithinbp:
Shouldn't we have like a common thread for everything? One that is not linked to any activity.
I don't know how this is implemented in Java, but in Python I once wrote a communication thread with which data is exchanged via signals and corresponding slots.
E.g., If I want to read the buffer, I would pass the necessary parameters to the communication handler, and also designate a callback function to which the communication handler will pass the buffer when it has finished handling. My main UI then picks up execution from the callback function.
But this approach requires all calls to go through the communication thread in order to avoid clashes

viveksb007:
Yes I thought about it too. To make a thread which has same lifetime as of UI thread and intercommunicate with each other. Haven't found a way to implement it in Android yet. Asyn task that @akarshan96 used is meant for small work max 4-5 sec. 
```
AsyncTask is designed to be a helper class around Thread and Handler and does not constitute a generic threading framework. AsyncTasks should ideally be used for short operations (a few seconds at the most.) If you need to keep threads running for long periods of time, it is highly recommended you use the various APIs provided by the java.util.concurrent package such as Executor, ThreadPoolExecutor and FutureTask.
```
For now we are creating a thread for single task and it gets killed after doing its task.
I would look for how to implement such architecture like its suggested to use ThreadPoolExecuter, FutureTask, etc. The clash happened in this case UI thread called for Version String and clashed with Initialisation thread. But I solved it here by Saving Version string in between when device is first connected and disconnected, So it doesn't call for version string again and again.

PR #217, opened 2017-06-22
Fixed : abrupt disconnections of device and prevent version read during initialisation

viveksb007:
Fixes issue #72 and #216 

Changes: 

- Abrupt disconnection is handled by receiving USB_DEVICE_DETACHED intent in BroadcastReceiver.
- prevents version read again and again by storing it as preference when device is connected and discarding it when its disconnected.   

[APK Link](https://drive.google.com/open?id=0B30IF3fyX4Q3M2xQM3hFVHBNVjg) for testing.

App crashes if device is removed during initialisation as initialisation is going on another thread ( bulkTransfer methods are called before broadcastReceiver can handle things ) and thread.stop() and thread.kill() methods are deprecated. Would look for something that prevents app crash and thread would be killed on its own after some time.

One commit from previous PR is reflected as I forgot to change branch.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -18,7 +18,8 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
-        <activity android:name=".activity.MainActivity"
+        <activity
+            android:name=".activity.MainActivity"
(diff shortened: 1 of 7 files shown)


viveksb007:
@asitava1998 could you open another issue for that?
And thanks for the review everyone.

Issue #234, opened 2017-06-26
allow negative numbers input in setting voltages and call methods on clicking Set button


viveksb007:
**Actual Behaviour**

can't set negative number through soft keyboard in Control Main.

**Expected Behaviour**

should be able to set negative numbers by soft keyboard.

**Steps to reproduce it**

Open Control Activity and try to input negative values in edit text.

**LogCat for the issue**

None

**Screenshots of the issue**

None

**Would you like to work on the issue?**

Yes.

viveksb007:
Fixed in #235

PR #235, opened 2017-06-26
negative numbers fixed and added callbacks

viveksb007:
Fixes issue #234 

Changes: allowed negative number input in edit text and added callbacks to set voltage and current.

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/adapters/ControlMainAdapter.java
+++ b/app/src/main/java/org/fossasia/pslab/adapters/ControlMainAdapter.java
@@ -1,6 +1,5 @@
 package org.fossasia.pslab.adapters;
 
-import android.content.Context;
 import android.support.v7.widget.CardView;
 import android.support.v7.widget.RecyclerView;
 import android.text.Editable;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/pslab/adapters/ControlMainAdapter.java
```diff
@@ -64,75 +70,63 @@ public void onBindViewHolder(final ViewHolder holder, int position) {
...
         final Button buttonControlMain2 = holder.buttonControlMain2;
-        Button buttonControlMain3 = holder.buttonControlMain2;
-
-        final SeekBar seekbarControlMain = holder.seekbarControlMain;
-        final EditText edittextControlMain = holder.edittextControlMain;
-
-        buttonControlMain1.setEnabled(false);
-        buttonControlMain2.setEnabled(false);
-
-        edittextControlMain.addTextChangedListener(new TextWatcher() {
```
asitava1998: I added this part because when the edittext boxes are empty and the buttons beside them are pressed, there is a crash. Please check it once.

viveksb007: I filled the values with 0, initially (assuming initial values of channel is 0) . So no crash there.

PR #238, opened 2017-06-27
Fixes #201 Adds functionality to Sensor QuickView

### LINKED ISSUE #201: Adding functionalities to Sensor QuickView
**Actual Behaviour**

As of now, there is just a UI for the Sensor QuickView. Functionalities to buttons and science methods haven't been used so far.

**Expected Behaviour**

The feature should enable the user to detect a sensor and get the raw data from that sensor.

**Steps to reproduce it**

Main menu -> Applications -> Sensor QuickView

**LogCat for the issue**

None

**Screenshots of the issue**

The textviews look odd as of now, but will be proper when filled with actual data.

Yes

asitava1998:
Fixes issue #201

Changes: 
* Added scienceLab methods and functions to buttons, spinners etc.
* As of now, no I2C sensors are available with me for testing. As soon as I receive my order of sensors, the code will be tested and further changes will be made.
* Some minor change in a few sensor files.
Screenshots for the change:

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/activity/SensorActivity.java
+++ b/app/src/main/java/org/fossasia/pslab/activity/SensorActivity.java
@@ -3,25 +3,320 @@ package org.fossasia.pslab.activity;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
 import android.support.v7.app.AppCompatActivity;
+import android.view.View;
+import android.widget.Button;
+import android.widget.Spinner;
+import android.widget.TextView;
 
 import org.fossasia.pslab.R;
 import org.fossasia.pslab.communication.ScienceLab;
+import org.fossasia.pslab.communication.peripherals.I2C;
+import org.fossasia.pslab.communication.peripherals.SPI;
+import org.fossasia.pslab.communication.sensors.ADS1115;
+import org.fossasia.pslab.communication.sensors.BH1750;
+import org.fossasia.pslab.communication.sensors.BMP180;
+import org.fossasia.pslab.communication.sensors.HMC5883L;
+import org.fossasia.pslab.communication.sensors.MLX90614;
+import org.fossasia.pslab.communication.sensors.MPU6050;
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/fossasia/pslab/activity/SensorActivity.java
```diff
@@ -3,25 +3,321 @@
...
+        buttonSensorGetRaw.setEnabled(false);
+
+        buttonSensorAutoscan.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                ArrayList<Integer> data = new ArrayList<Integer>();
+                ArrayList<String> datastr = new ArrayList<String>();
+                String datadisp = "";
+                try {
+                    i2c.init();
```
jithinbp: init does not need to be called here

asitava1998: Just the scan method should only be called, right?

jithinbp: try logging the results of I2C scan function

jithinbp: Yes. but init should not make a difference.
I think the problem with reading single bytes is probably causing issues here. Check the logs

asitava1998: I don't have any I2C sensors as of now to actually try out the code but I did order `MPU6050` and `HMC5883L` . I will test this soon.

jithinbp:
You don't a sensor to check the autoscan method though.
The onboard DAC is an I2C device with address 96 . It should be detected even if nothing is plugged in externally

asitava1998:
@jithinbp Then let me check that autoScan method then.

viveksb007:
@asitava1998 getAcknowledgement method is used during scanning. Are you getting any read error in the logs or its working fine?

asitava1998:
No errors as of now

Issue #243, opened 2017-06-27
bug in SPI setParameter function


viveksb007:
**Actual Behaviour**

incoming argument is sent to communicationHandler.

**Expected Behaviour**

modified argument should be sent.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

Yes

viveksb007:
Fixed in #245

Issue #242, opened 2017-06-27
bug in communicationHandler


viveksb007:
**Actual Behaviour**

CommunicationHandler is buggy, that's the reason some bytes transfer aren't successful.
For example : [here](https://github.com/fossasia/pslab-android/blob/master/app/src/main/java/org/fossasia/pslab/communication/CommunicationHandler.java#L129) bytesToBeRead is changes inside loop and condition depends on it too. 

**Expected Behaviour**

Solve bugs in CommunicationHandler.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

Yes

viveksb007:
Fixed in #245

Issue #240, opened 2017-06-27
Implement remaining capture methods in Oscilloscope UI


akarshan96:
**Actual Behaviour**

Currently, captureTraces and fetchTraces methods are used to fetch and plot the data from a single channel. 

**Expected Behaviour**

Extend functionality to fetch and plot data from multiple channels. For this, the following methods are required to be implemented in Oscilloscope UI -

- [x] captureTwo
- [x] captureFour

**Steps to reproduce it**

Menu Activity -> Applications -> Oscilloscope -> ChannelsParameter ->Select Multiple Checkboxes

**LogCat for the issue**

> None

**Screenshots of the issue**

> None

**Would you like to work on the issue?**

Yes

akarshan96:
fixed in #246

PR #245, opened 2017-06-28
fixed bug in communicationHandler and setParameters of SPI class

viveksb007:
Fixes issue #243 #242 

Now single byte can be read successfully.
Please test with this [apk](https://drive.google.com/open?id=0B30IF3fyX4Q3MkhuWm9zX2stS1U).


jithinbp:
Hi, 
there are some issues with the scope. since it's a WIP, i'm not sure which ones are issues
* One channel should be selected on startup. 'CH1' checkbox can be enabled by default
* None of the other channels are active yet, right?
* please check the timebase slider. only the 875uS value shows a sine wave. when I roughly double it to 2mS , the plot becomes a nearly straight line (I think the data scaling may be incorrect).
* When device is disconnect while the scope is running, it should return to the main activity. It returns to the main activity when i plugged it back in
have you tried checking the time taken to read the whole buffer for this sample size?

app/src/main/java/org/fossasia/pslab/activity/OscilloscopeActivity.java
```diff
@@ -426,15 +426,19 @@ protected Void doInBackground(String... params) {
                 analogInput = params[0];
                 //no. of samples and timegap still need to be determined
                 scienceLab.captureTraces(1, 1000, 10, analogInput, false, null);
+                try {
+                    Thread.sleep((long) (1000 * 1000 * 10 * 1e-6));
```
CloudyPadmal: This is `10` isn't it? :thinking:

viveksb007: Yes. Its just for testing to be clear of (samples * timeGap * 1e-3). Later it would be replaced with variables.

akarshan96:
@jithinbp currently timebase slider is not working. Will create an issue for it once over with #240. I will work on your suggestion to enable CH1 checkbox by default and App returns back to Main Activity when the device is disconnected.
Thanks

viveksb007:
@jithinbp This issue isn't related to scope. 
Still, I added a delay between captureTraces & fetch trace. And fixed communicationHandler Issue.
@akarshan96 Please open issues related to adding functionality to scope.

app/src/main/java/org/fossasia/pslab/activity/OscilloscopeActivity.java
CloudyPadmal: That's cool! :+1:

PR #246, opened 2017-06-28
Implemented captureTwo and captureFour methods in UI

akarshan96:
Partly Fixes issue #240 

Changes:
- Implemented captureTwo method in UI
- Implemented catureFour method in UI
- Fixed some capture methods
(There are read errors)
Screenshots for the change: 
video link - https://goo.gl/photos/hWonmBciNnmhZZ2P9

screenshot_2017-06-29-00-01-32-460_org fossasia pslab

screenshot_2017-06-29-00-16-23-202_org fossasia pslab

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/activity/OscilloscopeActivity.java
+++ b/app/src/main/java/org/fossasia/pslab/activity/OscilloscopeActivity.java
@@ -1,3 +1,4 @@
+
 package org.fossasia.pslab.activity;
 
 
@@ -36,6 +37,7 @@ import com.github.mikephil.charting.components.YAxis;
 import com.github.mikephil.charting.data.Entry;
 import com.github.mikephil.charting.data.LineData;
 import com.github.mikephil.charting.data.LineDataSet;
+import com.github.mikephil.charting.interfaces.datasets.ILineDataSet;
 
 import org.fossasia.pslab.communication.ScienceLab;
 import org.fossasia.pslab.fragment.ChannelParametersFragment;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
```diff
@@ -177,7 +177,7 @@ private void runInitSequence(Boolean loadCalibrationData) throws IOException {
             for (String temp : new String[]{"W1", "W2"}) {
                 loadEquation(temp, "sine");
             }
-            spi.setParameters(1, 7, 1, 0, null);
+            spi.setParameters(1, 7, 1, 0, -1);
```
viveksb007: Why this?
I changed it to null after modify spi setParameter function.

app/src/main/java/org/fossasia/pslab/activity/OscilloscopeActivity.java
```diff
@@ -103,7 +105,11 @@
     Fragment dataAnalysisFragment;
     Fragment xyPlotFragment;
     private final Object lock = new Object();
+    private final Object lock2 = new Object();
+    private final Object lock3 = new Object();
     private Task task;
+    private TaskTwo task2;
```
viveksb007: Please change the names to captureTwoTask ,etc.
And I think we can do away with single lock object. What do you think?

viveksb007:
@akarshan96 have you applied sine wave on both CH1 & CH2 in complete video duration?
And for the read error its due to timeout. When you try to read 1600 bytes it reads 400, sometimes 800. Need to go through the flow once again.

akarshan96:
@viveksb007 initially I applied the sine wave to CH1 then to CH2. About the timeout error, I doubled it but got the same results, I referred [this](https://stackoverflow.com/questions/14016943/android-bulktransfer-return-1-when-read-data-but-there-is-really-some-data-in-t) source. Let me try again fixing this. I have made the suggested changes.
Thanks

PR #252, opened 2017-06-29
Fixes #248 : App crash when multiple channels are selected

### LINKED ISSUE #248: Application crash in oscilloscope activity
**Actual Behaviour**

When a device is not connected, if I try to choose more than one channel from `Channel Parameters`, application crashes

**Expected Behaviour**

Application should not crash when multiple check-boxes are selected in oscilloscope

**Steps to reproduce it**

1. Open app
2. Go to `Applications`
3. Select `Oscilloscope`
4. Try selecting more than one checkbox out of `CH1`, `CH2`, `CH3` and `Microphone`

**LogCat for the issue**

```
06-29 09:43:07.000 32283-910/org.fossasia.pslab E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #2
                                                                  java.lang.RuntimeException: An error occured while executing doInBackground()
                                                                      at android.os.AsyncTask$3.done(AsyncTask.java:299)
                                                                      at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:352)
                                                                      at java.util.concurrent.FutureTask.setException(FutureTask.java:219)
                                                                      at java.util.concurrent.FutureTask.run(FutureTask.java:239)
                                                                      at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:230)
                                                                      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1080)
                                                                      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:573)
                                                                      at java.lang.Thread.run(Thread.java:856)
                                                                   Caused by: java.lang.IndexOutOfBoundsException: Invalid index 0, size is 0
                                                                      at java.util.ArrayList.throwIndexOutOfBoundsException(ArrayList.java:251)
                                                                      at java.util.ArrayList.get(ArrayList.java:304)
                                                                      at org.fossasia.pslab.communication.ScienceLab.fetchChannel(ScienceLab.java:897)
                                                                      at org.fossasia.pslab.communication.ScienceLab.fetchTrace(ScienceLab.java:867)
                                                                      at org.fossasia.pslab.activity.OscilloscopeActivity$Task.doInBackground(OscilloscopeActivity.java:429)
                                                                      at org.fossasia.pslab.activity.OscilloscopeActivity$Task.doInBackground(OscilloscopeActivity.java:419)
                                                                      at android.os.AsyncTask$2.call(AsyncTask.java:287)
                                                                      at java.util.concurrent.FutureTask.run(FutureTask.java:234)
                                                                      at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:230) 
                                                                      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1080) 
                                                                      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:573) 
                                                                      at java.lang.Thread.run(Thread.java:856) 

```

**Screenshots of the issue**

crash

Yes!

CloudyPadmal:
Fixes issue #248 

Changes: 
- Application tries to read from an array when there is no values in it when a device is not connected. Caught the exception in this fix

Screenshots for the change: 
crashfix

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
+++ b/app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
@@ -872,15 +872,11 @@ public class ScienceLab {
     }
 
     public HashMap<String, double[]> fetchTrace(int channelNumber) {
-        try {
-            this.fetchChannel(channelNumber);
-            HashMap<String, double[]> retData = new HashMap<>();
-            retData.put("x", this.aChannels.get(channelNumber - 1).getXAxis());
-            retData.put("y", this.aChannels.get(channelNumber - 1).getYAxis());
-            return retData;
-        } catch (IndexOutOfBoundsException e) {
-            return null;
-        }
+        this.fetchChannel(channelNumber);
+        HashMap<String, double[]> retData = new HashMap<>();
+        retData.put("x", this.aChannels.get(channelNumber - 1).getXAxis());
+        retData.put("y", this.aChannels.get(channelNumber - 1).getYAxis());
+        return retData;
     }
 
     public int[] oscilloscopeProgress() {
--- a/app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
+++ b/app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
@@ -872,11 +872,15 @@ public class ScienceLab {
     }
 
     public HashMap<String, double[]> fetchTrace(int channelNumber) {
-        this.fetchChannel(channelNumber);
-        HashMap<String, double[]> retData = new HashMap<>();
-        retData.put("x", this.aChannels.get(channelNumber - 1).getXAxis());
-        retData.put("y", this.aChannels.get(channelNumber - 1).getYAxis());
-        return retData;
+        try {
+            this.fetchChannel(channelNumber);
+            HashMap<String, double[]> retData = new HashMap<>();
+            retData.put("x", this.aChannels.get(channelNumber - 1).getXAxis());
+            retData.put("y", this.aChannels.get(channelNumber - 1).getYAxis());
+            return retData;
+        } catch (IndexOutOfBoundsException e) {
+            return null;
+        }
     }
 
     public int[] oscilloscopeProgress() {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
```diff
@@ -872,11 +872,15 @@ public void captureHighResolutionTraces(String channel, int samples, double time
...
-        retData.put("x", this.aChannels.get(channelNumber - 1).getXAxis());
-        retData.put("y", this.aChannels.get(channelNumber - 1).getYAxis());
-        return retData;
+        try {
+            this.fetchChannel(channelNumber);
+            HashMap<String, double[]> retData = new HashMap<>();
+            retData.put("x", this.aChannels.get(channelNumber - 1).getXAxis());
+            retData.put("y", this.aChannels.get(channelNumber - 1).getYAxis());
+            return retData;
+        } catch (IndexOutOfBoundsException e) {
```
viveksb007: why would this exception occur here?

CloudyPadmal: Hi @viveksb007 ! 
`this.aChannels` is empty when the device is not connected and throws the `IndexOutOfBoundException` when it tries to get values from it. Since there are three implementations calling functions to fetch values from `aChannels`, I caught the exception at this point and returned a null object so that in the background task it can be identified as an error and  proceed no further. Is that a good approach?

viveksb007: Well if the device is not connected there is no point in starting the background task rather than catching its exceptions.
You can prevent the call from Oscilloscope Activity by sciencelab.isConnected() and could you cross check that its updated when device is disconnected.
Thanks

CloudyPadmal:
@viveksb007 thanks for the tips! I added `scienceLab.isConnected()` check to the task initiation checkups. It fixes it without having to catch the exception :)

viveksb007:
@CloudyPadmal could you also for case when we have connected device, open Oscilloscope Activity and disconnect it and then select channel. If it crashes then, add one more check ``` sciencelab != null``` before sciencelab.isConnected() and change the order please in ```if``` condition, as once a false is encountered it doesn't check other arguments. So order should be ```sciencelab != null && scienceLab.isConnected() && booleans```

PR #253, opened 2017-06-30
Closes Oscilloscope Activity when device is disconnected

### LINKED ISSUE #247: Use LED icon in Oscilloscope to show the current status of the device
**Actual Behaviour**

There is no indication of current status of the device in Oscilloscope.

**Expected Behaviour**

A small green LED icon that flickers every time data is being read. It stops flickering and turn to red if the device is not longer connected.

**Steps to reproduce it**

Menu Drawer -> Applications -> Oscilloscope 

**LogCat for the issue**

> None

**Screenshots of the issue**

> None

Yes

akarshan96:
Fixes #247 

Changes: 
- Oscilloscope Activity closes when the device is disconnected.
- Reformatted code

Screenshots for the change: 
Video link - https://goo.gl/photos/v29Eho4HWxor5WkV8

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/receivers/USBDetachReceiver.java
+++ b/app/src/main/java/org/fossasia/pslab/receivers/USBDetachReceiver.java
@@ -1,5 +1,6 @@
 package org.fossasia.pslab.receivers;
 
+import android.app.ActivityManager;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
@@ -11,9 +12,14 @@ import android.widget.Toast;
 
 import org.fossasia.pslab.R;
 import org.fossasia.pslab.activity.MainActivity;
+import org.fossasia.pslab.activity.OscilloscopeActivity;
 import org.fossasia.pslab.fragment.HomeFragment;
 import org.fossasia.pslab.others.PreferenceManager;
 import org.fossasia.pslab.others.ScienceLabCommon;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 /**
  * Created by viveksb007 on 21/6/17.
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/pslab/receivers/USBDetachReceiver.java
```diff
@@ -39,20 +45,27 @@ public void onReceive(Context context, Intent intent) {
...
 
-                    if (activityContext != null) {
-                        MainActivity mainActivity = (MainActivity) activityContext;
-                        Fragment currentFragment = mainActivity.getSupportFragmentManager().findFragmentById(R.id.frame);
-                        if (currentFragment instanceof HomeFragment) {
-                            mainActivity.getSupportFragmentManager().beginTransaction().replace(R.id.frame, HomeFragment.newInstance(false, false)).commitAllowingStateLoss();
-                        }
+                    ArrayList<String> runningactivities = new ArrayList<String>();
+                    ActivityManager activityManager = (ActivityManager)context.getSystemService (Context.ACTIVITY_SERVICE);
+                    List<ActivityManager.RunningTaskInfo> services = activityManager.getRunningTasks(Integer.MAX_VALUE);
```
viveksb007: getRunningTasks method is deprecated in API level 21.
Didn't you saw that?

viveksb007:
And is this really required? I think when user is watching wave on oscilloscope and device is disconnected. It should just stop and user can go back or inspect wave plotted till the time device was connected.
Still of you want there are cleaner ways to do this.

akarshan96:
@viveksb007 it was suggested by Jithin in PR #245. Yeah, we can have a discussion on the fate of Activities when the device is unplugged. Once it's decided, will use another approach.

jithinbp:
@viveksb007, even if you don't want to exit the screen there should be some indication that the data is no longer being refreshed. Can we add setDisabled the widgets around the graph? Or show a disconnected label somewhere?

viveksb007:
@akarshan96 I think we can show a label on graph for telling if its actively updated with data. I saw something like ```Description label``` in MPAndroid Chart. Could you have a look?
And ```getRunningTasks``` has been deprecated because of security issue. 
Another way to do this is by making a class extending Application and implementing Activity callbacks.
I can push a commit for that, see if there is some label we can show user as ```STATUS```.

akarshan96:
@viveksb007 we can definitely use `Description label` to indicate the status. Let me work on it, I will share the screenshots for the same.

akarshan96:
@viveksb007 sharing screenshot of `Description label` indicating status. Plus we can do is to disable the widgets when the device is disconnected as suggested by @jithinbp. @asitava1998 @CloudyPadmal What say?

screenshot_1498935913

jithinbp:
I vote in favour of a small green LED icon that flickers every time data is being read. It will stop flickering and turn to red if device is not longer connected. Thoughts?

akarshan96:
@jithinbp great idea 👍

CloudyPadmal:
That's a good point @jithinbp ! :+1:

Issue #255, opened 2017-07-01
bug in interpretation of data


viveksb007:
**Actual Behaviour**

Currently its was assumed that MSB is read first.

**Expected Behaviour**

LSB is read first, so unpacking of data should be changed accordingly.  

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

Yes

Issue #256, opened 2017-07-01
wrong order of coefficients in Polynomial Function creation


viveksb007:
**Actual Behaviour**

currently order of coefficients is same as python comm lib

**Expected Behaviour**

In Polynomial Function by Apache Common Math, coefficients should be given in reverse order of what is in python comm lib for correct calculation of value at some data point.  

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

Yes

PR #257, opened 2017-07-01
Improved readback speed and fixed various bugs

### LINKED ISSUE #249: Read error during fetchTrace
**Actual Behaviour**

There is read error during fetchTrace. Most probably due to timeout. Due to which chart update on oscilloscope Activity is quite slow.

**Expected Behaviour**

There should be any read error and oscilloscope Activity shouldn't be laggy.

**Steps to reproduce it**

Open oscilloscope Activity and see chart update is quite slow.

Yes

viveksb007:
Fixes issue #255 #256 
Partially Fixes #249 

Changes: 
- fixed calibration data unpacking 
- byte order in making int from bytes
- fixed unsigned problem of byte ( In JAVA unsigned byte isn't supported, casted byte as ```int```, then ```ANDed``` with ```0xff```  )
- Fixed ```Polynomial Function``` Creation, unlike numpy Apache Common Math calculates value at a data point with reverse order of coefficient w.r.t ```numpy1D``` function.  

There are still readback error but now we can read most of sample data.

[Screencast](https://goo.gl/photos/ANicGk9EpygdKopD9) 
[APK Link](https://drive.google.com/open?id=0B30IF3fyX4Q3bXB4SVJnQ0dLems)

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/activity/OscilloscopeActivity.java
+++ b/app/src/main/java/org/fossasia/pslab/activity/OscilloscopeActivity.java
@@ -137,8 +137,8 @@ public class OscilloscopeActivity extends AppCompatActivity implements
         y1 = mChart.getAxisLeft();
         y2 = mChart.getAxisRight();
 
-        int freq = scienceLab.setSine1(3000);
-        Log.v("SIN Fre", "" + freq);
+        //int freq = scienceLab.setSine1(3000);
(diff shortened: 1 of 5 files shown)


asitava1998:
Why is there an occasional distortion in the sine wave?

viveksb007:
there are still read-back errors, so sample data isn't uniform as it should be so the distortion is due to that.

Issue #258, opened 2017-07-01
version string is read when switching between screens to HomeFragment


viveksb007:
**Actual Behaviour**

Device version string is read when switch between screens to HomeFragment.

**Expected Behaviour**

Version should be read once after device is connected. It doesn't take much time but sometimes after we go back from oscilloscope, version read clashes with thread that was reading captureData. Plus there is no point reading version data again and again. 

**Steps to reproduce it**

Switch between screens and read logs.

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

Yes

PR #259, opened 2017-07-01
fixed version read between screens to HomeFragment

viveksb007:
Fixes issue #258 

Changes: Stored Version in ```PacketHandler Version``` variable and read from it during the time app is opened. Version is read only once when device is connected.

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/fragment/HomeFragment.java
+++ b/app/src/main/java/org/fossasia/pslab/fragment/HomeFragment.java
@@ -11,9 +11,9 @@ import android.widget.ImageView;
 import android.widget.TextView;
 
 import org.fossasia.pslab.R;
+import org.fossasia.pslab.communication.PacketHandler;
 import org.fossasia.pslab.communication.ScienceLab;
 import org.fossasia.pslab.others.InitializationVariable;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/pslab/fragment/HomeFragment.java
```diff
@@ -87,12 +87,10 @@ public void onViewCreated(View view, @Nullable Bundle savedInstanceState) {
                 public void run() {
                     try {
 
-                        String version;
-                        PreferenceManager preferenceManager = new PreferenceManager(getActivity());
-                        if ("none".equals(preferenceManager.getVersion())) {
+                        String version = PacketHandler.version;
+                        if ("".equals(version)) {
```
CloudyPadmal: Can't we use `TextUtils.isEmpty(version)`? :)

viveksb007: [TextUtils.isEmpty](https://developer.android.com/reference/android/text/TextUtils.html#isEmpty(java.lang.CharSequence)), checks if a string is null or 0-length. It can't be null as I have initialised it with 0-length. So I am checking for 0-length only.

PR #271, opened 2017-07-04
Fixes #262 completes remaining portion of Sensor Activity

### LINKED ISSUE #262: Complete remaining portion of Sensor Activity
**Actual Behaviour**

Presently the `AutoScan` button of Sensor QuickView works and the results of AutoScan are displayed. However, the list of available sensors are not reflected under `Select Sensor` and as a result `Get raw` is not functional.

**Expected Behaviour**

Both the `Select Sensor` spinner and `Get Raw` button must be available.

**Steps to reproduce it**

Main Menu -> Applications -> Sensor Quickview -> Click on `AutoScan` -> The result will be displayed in the test-box below but the spinner & button below it are not functional.

**LogCat for the issue**

None

**Screenshots of the issue**

None

Yes

asitava1998:
Fixes issue #262 

Changes:
* Detected sensors will be added in the spinner.
* Included `scienceLab.isConnected()` at relevant points.
* Map with name of sensors and keys being their unique addresses. 

[APK](https://drive.google.com/file/d/0B3F6wGD_1dcZaGxuTXQxcFhKR2M/view?usp=sharing) for testing. Maybe @jithinbp can test the apk. 

Screenshots for the change:


jithinbp:
Cool!
Few thoughts.
* Is it possible to add a checkbox to getRaw, such that the values are continuously updated?
* A slightly more formatted output so that we can distinguish the numbers

otherwise good enough for checking sensor library functionalities. I see both your sensors are working.

asitava1998:
I will work on formatting the outputs for each sensor and I was thinking of implementing sensor specific UIs. When we select a sensor from the spinner, it should take you to the specific UI for that sensor.

viveksb007:
@asitava1998 I think there is a bit of confusion. @jithinbp asked for a developer only option for testing the sensors. So what you did till now should belong to Data Sensor Logger. 
User would go to Sensor Quick View, scan for sensors and a list of scanned sensors would be presented (multiple sensors can be connected simultaneously) . User would choose any sensor from the list and data corresponding to that sensor would be presented with specific UI like for MPU6050, UI showing X,Y,Z, etc. With an option to visualise them in graph and other stuff.
To test any new sensor we can immediately code in Data Sensor Logger with worrying about the UI.
And data logging should be continuous (you might be planning that with UI). 
@jithinbp that's the roadmap right?

asitava1998:
@viveksb007 No, there is no confusion regarding that. Here is the thing.
* The work done so far is part of Sensor QuickView and not of Data Sensor Logger. The current UI was just a `temporary` one to see if the sensor libraries are correct or not (Two of them which i tested had mistakes) and to get the raw value from the sensors.
* Here is the work on [mock-ups](https://app.moqups.com/asitavasarkar/Wwr6F3xmaB/view). Once these are approved ,the Get Raw function will be taken out from the UI, and on selecting the sensor we will land on the corresponding fragment for that sensor.
* If you saw the mock-ups, the raw values are displayed and this where Get Raw is needed and there is a provision for plots in there as well.
* Please view the Sensor QuickView in the desktop app. Our implementation is almost similar to it.

viveksb007:
Got your point. Nice mockups 👍 
What is Data Sensor Logger for?

asitava1998:
@viveksb007 Sensor Data Logger in the desktop app was meant for recording a set of data and plotting them and Sensor QuickView was for displaying the values only (plotting not included).

I feel it will be better if we do both of them together under a common heading as shown in the mock-ups like Science Journal. The name Sensor QuickView should still be good even if we do both of them together.

viveksb007:
Exactly my point functionality is overlapping between the two.
Now We still need Data Sensor Logger to test new sensors without breaking functionality of Sensor Quick View.
Just to be clear I am repeating the flow for Sensor Quick View.
User Scans for attached sensors. A **Listview** is presented not **Spinner**. And after clicking on **listItem** i.e some sensor, a fragment comes over with that sensor specific UI exposing functionality like showing realtime sensor data, option for plotting and logging or any other add on.

jithinbp:
Yes. I believe the UIs can be combined. Like the science journal app has both readings and plots if required

app/src/main/java/org/fossasia/pslab/communication/sensors/HMC5883L.java
```diff
@@ -30,44 +30,44 @@
...
     private I2C i2c;
 
     public HMC5883L(I2C i2c) throws IOException {
         this.i2c = i2c;
 
-        params.put("init", new ArrayList(Arrays.asList("Now")));
-        params.put("setSamplesToAverage", new ArrayList(Arrays.asList(samplesToAverageChoices)));
-        params.put("setDataOutputRate", new ArrayList(Arrays.asList(dataOutputRateChoices)));
-        params.put("setGain", new ArrayList(Arrays.asList(gainChoices)));
+        params.put("init", null);
```
viveksb007: This params dictionary is used in python to generate GUI dynamically.
If we make sensor specific UI then we won't be needing this.

asitava1998: Fixed :+1:

asitava1998:
Either way we will be editing all sensor files, so we will change at that time.

Issue #260, opened 2017-07-01
Implement functionalities in Oscilloscope UI


akarshan96:
**Actual Behaviour**

Some functionalities haven't been implemented in Oscilloscope UI yet

**Expected Behaviour**

Implement the following in Oscilloscope UI

- [x] Timebase 
- [x] Trigger
- [x] Curve Fittings
- [x] Fourier Transform

**Steps to reproduce it**

> None

**LogCat for the issue**

> None

**Screenshots of the issue**

> None

**Would you like to work on the issue?**

Yes

viveksb007:
@akarshan96 If this issue is solved. Please close it and open other issue relevant to what's remaining.

PR #282, opened 2017-07-06
Implements timebase in Oscilloscope UI

akarshan96:
Partly Fixes issue #260 

Changes: 
Timebase seek bar is functional

[apk link ](https://drive.google.com/open?id=0B423XEmWL9vKNlc3YzhMdlhOU1U)
*W1 produces sine wave

Screenshots for the change: 

screenshot_2017-07-06-01-53-59-940_org fossasia pslab

**Here, alternate subsections of 200 data points is reading wrong data**


app/src/main/java/org/fossasia/pslab/activity/OscilloscopeActivity.java
```diff
@@ -400,7 +401,6 @@ public void chartInit() {
         mChart.setPinchZoom(true);
         mChart.setScaleYEnabled(false);
         mChart.setBackgroundColor(Color.BLACK);
-        mChart.getDescription().setEnabled(false);
```
viveksb007: why this?

viveksb007:
@akarshan96 I tested the apk. The amplitude of wave is also quite low.
And **alternate subsections of 200 data points is reading wrong data**, there are packet drops randomly right?

akarshan96:
@viveksb007 Yes, random packet drops are causing this.

viveksb007:
@akarshan96 I don't think its random I am getting distortion at same time instants in all my tests.
screenshot_2017-07-07-16-41-28-125_org fossasia pslab

Even if its random that distorted section should move when triggered but its not.
Could you dig some more into whats the issue here?
Meanwhile I would also look for what can be possible issue.
@jithinbp any views as to why this can happen?

viveksb007:
@akarshan96 any updates on this? or should we merge and you would follow later?

akarshan96:
@viveksb007 merge it, I will follow it later.

PR #293, opened 2017-07-09
Added LED icon in Oscilloscope UI

### LINKED ISSUE #247: Use LED icon in Oscilloscope to show the current status of the device

akarshan96:
Fixes #247 
Changes: 
- Added LED icon which goes from red to green when data is read.
- Reformatted code

Apk:
[app-debug.apk.zip](https://github.com/fossasia/pslab-android/files/1133964/app-debug.apk.zip)

Screenshots for the change: 

screen shot 2017-07-10 at 12 12 37 am

screen shot 2017-07-10 at 12 12 21 am


app/src/main/java/org/fossasia/pslab/activity/OscilloscopeActivity.java
```diff
@@ -271,6 +273,18 @@ public void run() {
                             }
                         }
                     }
+
+                    if (!scienceLab.isConnected() || (!isCH1Selected && !isCH2Selected && !isCH3Selected && !isMICSelected)) {
```
CloudyPadmal: What if `!scienceLab.isConnected() &&` is used instead of `!scienceLab.isConnected() ||`? :)

viveksb007: Any one of them is sufficient for icon to be green. suppose device is connected and no channel is selected then also icon should be red as no data is refreshed.

CloudyPadmal: Ah right! :+1: Thanks @viveksb007 for the info!

Issue #292, opened 2017-07-09
Create UI for performing experiment listed in Saved Experiment Fragment


viveksb007:
**Actual Behaviour**

List of Experiment is shown in Saved Experiment Fragment.

**Expected Behaviour**

- [x] On clicking any experiment from that list corresponding Experiment Fragment/Activity should open.
- [x] Create layout for experiment doc. ( Would try putting pslab-desktop-app docs in local directory and show them in web-view ) (mapped Diode I-V, Zener I-V experiment )
- [x] Create layout for experiment setup ( this varies from experiment to experiment) would make for some 2-3 basic experiment for now. (Setup for Zener I-V created)

**Steps to reproduce it**

NavDrawer->Saved Experiment

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

Yes

viveksb007:
@jithinbp @akarshan96 @asitava1998 @CloudyPadmal 
Could you guys have a look and give your feedback/suggestions on layout?
Screencast Link : https://goo.gl/photos/LKLXjz9QkVuuFZnY7
 
Lower part of first image is completely different layout which would vary with experiment. In upper part only Image and Title would vary. Screencast has shown this better.
On clicking play button, one fragment would come over which would have view components to perform experiment.

<table>
<tr>
<td><img src="https://user-images.githubusercontent.com/12713808/28028365-e94efe84-65b9-11e7-862b-561cacbf82c2.png"></td>
<td><img src="https://user-images.githubusercontent.com/12713808/28028364-e94b8588-65b9-11e7-9db1-71fa6f2c27cb.png"></td>
</tr>
</table>

jithinbp:
Cool. It feels quite elegant.

I have questions/suggestions:
* When you hit the play button, is there going to be an easy way to switch between the doc screen and the experiment activity without having to reconfigure the parameters?  It would be nice if the two would work together with no need to use the 'back' button.
* In the main screen subtitles, we can maybe do away with the 'Experiments related to ...' text . It's probably obvious
* Can you reuse the docs from pslab-desktop-apps?  They use pictures from designated subfolders, so all you have to do is replace the 'screenshots' folder with app screenshots , and recompile the static site. Will save effort on the documentation front since the text, schematics and photographs are reusable.


thanks

viveksb007:
[...]

PR #303, opened 2017-07-12
Basic Perform Experiment UI done 

viveksb007:
Fixes issue #292 

Changes:
- Added Activity UI to perform experiment.

There is some unnecessary text in Experiment doc and Experiment format xml files. I would replace them with their actual layout for which I would open necessary issues.

To do (Would open another issue for this) : 
- Finalise Experiment doc format, probably use pslab-desktop-app docs using webView (need to try that)
- UI for Experiment Setup  

Screencast : https://goo.gl/photos/uACDUpT9emjeYHRz8

Screenshots for the change: 
<table>
<tr>
<td></td>
<td></td>
</tr>
</table>


app/src/main/java/org/fossasia/pslab/adapters/PerformExperimentAdapter.java
```diff
@@ -0,0 +1,45 @@
...
+    private final int PAGE_COUNT = 2;
+    private String[] tabTitle = new String[]{"Experiment Doc", "Experiment Setup"};
+
+    public PerformExperimentAdapter(FragmentManager fragmentManager) {
+        super(fragmentManager);
+    }
+
+    @Override
+    public Fragment getItem(int position) {
+        switch (position) {
```
CloudyPadmal: This can be put in an `if else` block right? :)

```java
if (position == 1) {
    return ExperimentSetupFragment.newInstance();
} else {
    return ExperimentDocFragment.newInstance();
}
```

viveksb007: Not currently, might be there in future like previous experiment trials or something.

CloudyPadmal: Ahh! Then that will be fine! :)

Issue #294, opened 2017-07-10
Implement feature to read and write buffer data from Audio Jack


viveksb007:
**Actual Behaviour**

Currently no support for Mic Audio Jack Hack discussed in [this doc](https://docs.google.com/document/d/1plDelUJ1EwQj5P7ZP0R61cQawPIn-qdFZlBynvKM4wM/edit?usp=sharing)

**Expected Behaviour**

Implement Using Audio Jack as Mic-In,etc functionality discussed in doc. Data from Mic In can be plotted on Oscilloscope. Other uses are in doc. 
- [x] Add AudioJack class to initialise AudioRecord and AudioTrack objects and create public methods to read and write buffer.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

Yes

viveksb007:
Fixed in #304

PR #304, opened 2017-07-12
started implementing using audioJack as probe and wavegenerator

viveksb007:
partially fixes issue #294 

Changes: 
- AudioJack class would initialise AudioRecord and AudioTrack objects and provide function to read and write buffer.

Need to test whether buffer read & write would work as it is or need to shift them to another thread. 

Screenshots for the change: NA


viveksb007:
I have just added methods for now. Would update with the screenshots as soon as I map it to oscilloscope as test it. Thanks for the link.

Issue #299, opened 2017-07-12
Edit sensor files to remove params


asitava1998:
**Actual Behaviour**

The sensor files have an ArrayList/HashMap `params` which was added while porting .

**Expected Behaviour**

`params` is not needed in Android since it was used in python library for the GUI.

**Steps to reproduce it**

pslab-android -> app -> src -> main -> java -> org -> fossasia -> pslab -> communication -> sensors

**LogCat for the issue**

None

**Screenshots of the issue**

None

**Would you like to work on the issue?**

Yes

PR #311, opened 2017-07-14
Removed params and edited sensor files

asitava1998:
Fixes issue #299 

Changes: 
* Removed params from sensor files which is not necessary for Android and in some cases replaced fields with corresponding ArrayLists.
* Some changes might be due to reformatting of code.

Screenshots for the change: 
None


viveksb007:
Could you have a look om codacy warnings. I think those unsed variables are also GUI related. Could you cross check it and remove if not required?
Thanks

asitava1998:
@viveksb007 There are some features which have not been implemented in the sensor files as Jithin also mentioned in the channel, we may implement them in the future. Better to leave these as they are for now.

PR #331, opened 2017-07-18
Removed working label from LogicAnalyzer

CloudyPadmal:
Changes: 
- Logic Analyzer is still being implemented. So removed the working label from the readme file


viveksb007:
@CloudyPadmal working means we are currently working on it. And established means its completely implemented.

PR #341, opened 2017-07-19
Created layout and Fragment files for SHT21 and TSL2561

asitava1998:
Fixes issue #334 

Changes: 
* Created layout files for `SHT21` and `TSL2561`.
Changes remaining:
* Create fragment files for the above

Screenshots for the change: 

I don't have TSL2561 for testing currently. However, all the sensors have the same errors in the values displayed. What can be the possible reason?


viveksb007:
@asitava1998 is there any read error in logs when you read raw values from sensor?

asitava1998:
The code works alright. I don't get any read error in the logs when I run it. I am still trying to understand where the error possibly is. I am going through the entire communication directory once again.
I don't think the error is in the individual sensor files as I don't think all of them could be ported in a wrong manner, the problem lies else where.

app/src/main/res/values/strings.xml
```diff
@@ -145,6 +147,8 @@
     <string name="gy">Gy</string>
     <string name="gz">Gz</string>
     <string name="temp">Temp</string>
+    <string name="luminosity">Luminosity</string>
```
viveksb007: ```luminosity``` key is repeated resulting in build failing. Please resolve it and ensure build is successful.

asitava1998: Fixed the build error. :+1:

PR #348, opened 2017-07-20
Fixes #346: LED indicator for Logic Analyzer

### LINKED ISSUE #346: There is no indication of the device status in Logic Analyzer
**Actual Behaviour**

User is not provided with an indication whether the device is connected or not

**Expected Behaviour**

As in the Oscilloscope, Logic Analyzer should support a feature like displaying the device status whether it is connected or not

**Steps to reproduce it**

1. Open app
2. Go to `Application`
3. Go to `Logic Analyzer` and observe

**LogCat for the issue**

> Not available

**Screenshots of the issue**

> Not available

Yes

CloudyPadmal:
Fixes issue #346 

Changes: 
- Adopted implementation from Oscilloscope done by Akarshan! (Special thanks :) )
- LED will change from green to red when device is unplugged
- LED will remain red when device is not plugged

Screenshots for the change: 
[Screencast](https://drive.google.com/open?id=0B5su2_aUn8g7Q1lPNEs1bmpUMXc)

APK for testing:
[app-debug.apk.zip](https://github.com/fossasia/pslab-android/files/1161126/app-debug.apk.zip)

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
+++ b/app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
@@ -99,6 +99,11 @@ public class LALogicLinesFragment extends Fragment {
                             });
                         }
                         // TODO : Perform Logic Analyzer related functions
+                        // Colors for Data sets
+                        // - CH1 - (Color.BLUE);
+                        // - CH2 - (Color.GREEN);
+                        // - CH3 - (Color.RED);
+                        // - CH4 - (Color.YELLOW);
                     } else {
                         if (!String.valueOf(ledImageView.getTag()).equals("red")) {
                             getActivity().runOnUiThread(new Runnable() {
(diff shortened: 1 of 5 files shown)


viveksb007:
@CloudyPadmal In oscilloscope activity, indicator tells when data is fetched from device i.e when screen is refreshed with new data. We haven't got the minimum functionality working yet. Here indicator is green if device is connected with no relevance of data.
I think we should disable background grid lines if we are shifting to black and white colour configuration.
And could you try ```FrameLayout``` instead of ```RelativeLayout```, in frame layout you can overlay that image on chart corner. Check if that works.
Thanks

CloudyPadmal:
Thanks for the review @viveksb007 ! :+1: Just like in the oscilloscope activity, in Logic analyzer activity, data is fetched as the linechart fragment is opened. I have added a `TODO` comment in the code where the background task needs to be called to make logic analyzer functionalities. I have followed a similar procedure as Akarshan has implemented in oscilloscope without trying to reinvent a new method.
Grid lines makes readings much easier, don't they? Also would a fragment layout make any difference as in oscilloscope activity also it is inside a relative layout?

viveksb007:
yes grid lines definitely aid in reading timestamp, both were of same color, so I suggested to avoid confusion. ```FrameLayout``` layout would allow one view over another, effectively rendering status image over chart corner.

CloudyPadmal:
Hi @viveksb007 ! I tried the frame layout instead of relative layout. We can set data set colors like the colors in the oscilloscope. Is it okay now? :) Much appreciated!
[Screencast](https://drive.google.com/file/d/0B5su2_aUn8g7eVRGVnFaS0VYRkU/view?usp=sharing), [APK](https://github.com/fossasia/pslab-android/files/1164421/app-debug.apk.zip)

Issue #364, opened 2017-07-23
add SensorDataLogger activity 


viveksb007:
Add sensorDataLogger activity and asking user for WRITE_STORAGE permission if app doesn't have.

Parent Issue #363

PR #365, opened 2017-07-23
Basic sensorDataLogger activity created 

viveksb007:
Fixes issue #364 

Changes:
- boiler plate code for activity added
- asks user for WRITE_STORAGE permission to log data


viveksb007:
@CloudyPadmal I saw the library for permission there is also [easypermissions](https://github.com/googlesamples/easypermissions) by google but we don't need much permissions plus it's almost same amount of code to do it yourself. So I did it on my own rather than using a library.

Issue #339, opened 2017-07-18
Add more functionality to Sensor fragments


akarshan96:
**Actual Behaviour**

Currently, Sensor fragments don't have all the functionalities.

**Expected Behaviour**

Sensor fragments should have

- [ ] Pause/record button
- [ ] Time gap between each data point option
- [ ] No. of data points to be recorded option

**Steps to reproduce it**

Menu Drawer -> Applications -> Sensor QuickView -> HMC5883L 

**LogCat for the issue**

> None

**Screenshots of the issue**

> None

**Would you like to work on the issue?**

Yes

akarshan96:
@jithinbp @asitava1998 @viveksb007 @CloudyPadmal, I am sharing a mockup for control bar in Sensor activity. I have added 

1. a common play and pause button. 
2. a seek-bar to determine the time gap between each data point capture.
3. an edit-text box in which no. of samples to be captured can be entered.
4. a checkbox when checked will capture data indefinitely. This will disable the edit-text box.

Please review it and give suggestions.

![screen shot 2017-07-23 at 8 52 06 pm](https://user-images.githubusercontent.com/26146675/28500980-fe326f44-6fef-11e7-9f79-1d177b306b25.png)

viveksb007:
@akarshan96 Looks great. Instead of using TextView and EditText for taking Number of samples use ```TextInputEditText``` wrapped inside ```TextInputLayout```. Refer [this](https://github.com/fossasia/pslab-android/blob/master/app/src/main/res/layout/diode_configure_dialog.xml#L17).

viveksb007:
@akarshan96 Please update the Issue description about what has been achieved and if anything is remaining, please open a sub-issue.

PR #372, opened 2017-07-26
Created control dock for Sensor Activity

akarshan96:
Partly fixes issue #339 

Changes: 
Added a control dock in sensor activity.

* Will make the dock functional in future PR's
Screenshots for the change: 

screen shot 2017-07-26 at 11 18 12 am


viveksb007:
@akarshan96 is play/pause simultaneously on the image?
Transitioning between play/pause would be great i.e only a play image which changes to pause on clicking something like a music player. What do you think?

akarshan96:
@viveksb007 earlier there was no transition but now I have modified the PR. Sharing the gif.

videotogif_2017 07 26_20 04 45

Issue #380, opened 2017-07-28
create a MathUtils class to add frequently used Math functions 


viveksb007:
**Actual Behaviour**

Currently functions like map(), etc which are not available have to be written in each file.

**Expected Behaviour**

Create a MathUtil class which can have frequently used math functions like map(), linspace(), etc which are provided in numpy are ported in Java.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

Yes

Issue #378, opened 2017-07-27
add functionality to experiment setups created


viveksb007:
**Actual Behaviour**

currently experiment setups are bare layout without functionality.

**Expected Behaviour**

Experiment setups should be functional to perform experiment.

Functionality for
- [x] Diode I-V
- [x] BJT related experiments
- [x] Ohms law

**Steps to reproduce it**

Saved Experiments -> Any experiment setup with layout except zener I-V.

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

Yes

PR #381, opened 2017-07-28
Added experiment setup functionality

viveksb007:
Partially Fixes issue #378 
Partially Fixes issue #380 

Changes:

Added experiment functionality 
- [x] diode I-V (reusing Zener I-V code as both are identical)
- [x] Ohms Law
- [x] transistor CB

Shifted **map()** function to MathUtils and modified its usages.Would gradually shift functions like **linspace()**,etc in numpy whose implementation code is repeated quite a lot of times.  

Screenshots for the change: 
Ohms experiment setup screencast : https://photos.app.goo.gl/YPXiO7i1wwyBd8uF2
screenshot_2017-07-29-16-58-32-457_org fossasia pslab

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/activity/OscilloscopeActivity.java
+++ b/app/src/main/java/org/fossasia/pslab/activity/OscilloscopeActivity.java
@@ -44,9 +44,10 @@ import org.fossasia.pslab.others.ScienceLabCommon;
 import org.fossasia.pslab.R;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+
+import static org.fossasia.pslab.others.MathUtils.map;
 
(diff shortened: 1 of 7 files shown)


viveksb007:
@CloudyPadmal MathUtils has definition of self-implemented functions which are frequently used. `DecimalFormat("0.0000")` doesn't qualify to be put there.

CloudyPadmal:
@viveksb007 I thought since the pattern 0.0000 is used in control UI as well as experimentals UIs, if it is set somewhere we can use it as a standard way of representing decimals with 4 decimal places! Anyways then it is just fine! :+1:

viveksb007:
@CloudyPadmal its a good idea to declare standards somewhere but I think MathUtils isn't suitable for that. 
Thanks for the review guys.

Issue #379, opened 2017-07-27
Modify the Sensor Activity Layout


asitava1998:
**Actual Behaviour**
The present layout is non-intuitive and just contains an `AutoScan` button, a Textbox and of whitespace as the ListView is displayed when sensors are detected.

**Expected Behaviour**
* In addition to `AutoScan`, a feature to automatically detect any connected sensors when Sensor QuickView is opened.
* Modify the style of text displayed in the textbox.

**Steps to reproduce it**
Main Menu -> Applications -> Sensor QuickView

**LogCat for the issue**
None

**Screenshots of the issue**
<img src="https://user-images.githubusercontent.com/19410364/28672034-26ac3c2a-72fc-11e7-9881-afc20e969a04.png" height="640" width="360">

**Would you like to work on the issue?**
Yes

viveksb007:
Fixed in #389

PR #389, opened 2017-07-30
Fixes sensor activity layout

asitava1998:
Fixes issue #379 

Changes: 
* The sensors detected are loaded as the fragment opens.
* Minor changes in text sizes, style, alignment etc.

Screenshots for the change:


app/src/main/java/org/fossasia/pslab/sensorfragment/SensorFragmentMain.java
```diff
@@ -64,35 +62,63 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
                     try {
                         data = i2c.scan(null);
                     } catch (IOException e) {
                         e.printStackTrace();
                     }
+
                     if (data != null) {
                         for (Integer myInt : data) {
-                            if (sensorAddr.get(myInt) != null) {
+                            if (sensorAddr.get(myInt) != null && myInt != null) {
```
viveksb007: If ```myInt``` is null, the current expression would first try to find ```null``` in map.
Its should be ```myInt != null && sensorAddr.get(myInt)```

app/src/main/java/org/fossasia/pslab/sensorfragment/SensorFragmentMain.java
```diff
@@ -138,8 +164,14 @@ public void onItemClick(AdapterView<?> parent, View view, int position, long id)
             }
 
         });
-
         return view;
+    }
+
+    private void setText (String text) {
```
viveksb007: Is there any specific reason for creating a function ```setText``` instead of using ```tv.setText```?
Same for ```setAdapter``` method below.

PR #412, opened 2017-08-02
Fixes #380 added linspace() to MathUtils

### LINKED ISSUE #380: create a MathUtils class to add frequently used Math functions

viveksb007:
Fixes issue #380 

Changes:
- Added implementation of linspace() from numpy to MathUtils Java.

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
+++ b/app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
@@ -783,16 +783,6 @@ public class ScienceLab {
                 this.buffer[i] = (listData.get(i * 2) | (listData.get(i * 2 + 1) << 8));
             }
 
-            /* Implementation moved to MathUtils
-
-            ArrayList<Double> timeBase = new ArrayList<>();
-            double factor = timeGap * (samples - 1) / samples;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/pslab/communication/ScienceLab.java
```diff
@@ -781,13 +782,18 @@ else if (args.contains("READ_CAP")) {
             for (int i = 0; i < samples; i++) {
                 this.buffer[i] = (listData.get(i * 2) | (listData.get(i * 2 + 1) << 8));
             }
+
+            /* Implementation moved to MathUtils
+
```
CloudyPadmal: Can't we remove this commented out code?

viveksb007: Yeah we can. Just kept it for reference to review the implementation.

Issue #409, opened 2017-08-02
Create layout and fragment for MPU925x


asitava1998:
**Actual Behaviour**
Currently there is no layout and fragment for MPU925x

**Expected Behaviour**
There should support for MPU925x in PSLab. 

**Steps to reproduce it**
None

**LogCat for the issue**
None

**Screenshots of the issue**
None

**Would you like to work on the issue?**
Yes

PR #417, opened 2017-08-03
Creates layout and fragment for MPU925X

asitava1998:
Fixes issue #409 

Changes: 
* Creates layout and fragment for MPU925X.

Issue:
* MPU6050 and MPU925X share the same address, distinguishing them at the device level is not possible. (It's possible for Arduino, but I don't know about PSLab).
* if it is not possible, we have to rely solely on the user to select the right sensor. This can be taken up as a separate issue.

Screenshots for the change:


akarshan96:
@asitava1998 can you see what is causing the build to fail. Thanks.

asitava1998:
@akarshan96 I updated the branch and then the build failed. I am looking into it.

asitava1998:
@akarshan96 @viveksb007 @CloudyPadmal Can we speed up the review process? The files in this PR are also present in other PRs and when those are merged, it leads to merge conflicts here.

app/src/main/java/org/fossasia/pslab/sensorfragment/SensorFragmentMPU925X.java
```diff
@@ -0,0 +1,323 @@
...
+
+            dataset1.setColor(Color.BLUE);
+            dataSet2.setColor(Color.GREEN);
+            dataSet3.setColor(Color.RED);
+
+            dataset4.setColor(Color.BLUE);
+            dataSet5.setColor(Color.GREEN);
+            dataSet6.setColor(Color.RED);
+
+            dataset1.setDrawCircles(true);
```
viveksb007: ```drawCircles``` are ```true``` by default, if you want to not draw circle at data point then you set it to ```false```.

CloudyPadmal:
Looks OK! :) Once the requested changes by Vivek is made I'll approve this! :+1:

app/src/main/java/org/fossasia/pslab/sensorfragment/SensorFragmentMPU925X.java
asitava1998: Changes done

viveksb007: What have you changed?
```dataset.setDrawCircles(true)``` doesn't make any difference, it true by default.

PR #426, opened 2017-08-04
Fixes #420 : Checks final voltage and initial voltage

### LINKED ISSUE #420: Current experiments do not check initial and final value differences
**Actual Behaviour**

In the current implementations of saved experiments, do not validate to see if the initial voltages are lesser or greater (BJT (less) and FET (great))

**Expected Behaviour**

Initial values should be matched with final values. Otherwise they will result in a negative step size which will cause an infinity for loop

**Steps to reproduce it**

1. Open app
2. Go to `Saved Experiments`
3. Choose `BJT CB Characteristics` experiment
4. Set initial value higher than final value and start experiment to observe

[Result could decrement for loop exiting variable](https://github.com/fossasia/pslab-android/blob/master/app/src/main/java/org/fossasia/pslab/experimentsetup/TransistorCBSetup.java#L148)

**LogCat for the issue**

> Not available

**Screenshots of the issue**

> Not available

Yes! I will implement a common method to overcome this as I am implementing FET related experiments at the moment.

CloudyPadmal:
Fixes issue #420 

Changes: 
- Added a check to final voltage values when they are set to function parameters

Screenshots for the change: 

fix

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/experimentsetup/NFETOutputCharacteristicsExperiment.java
+++ b/app/src/main/java/org/fossasia/pslab/experimentsetup/NFETOutputCharacteristicsExperiment.java
@@ -44,6 +44,7 @@ public class NFETOutputCharacteristicsExperiment extends Fragment {
      ***********************************************************************************************/
 
     private static final String ERROR_MESSAGE = "Invalid Value";
+    private static final String INVALID_VALUE = "Voltage value too low";
     private LineChart outputChart;
     private float initialVoltage;
     private float finalVoltage;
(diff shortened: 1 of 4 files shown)


CloudyPadmal:
APK for testing:

[app-debug.apk.zip](https://github.com/fossasia/pslab-android/files/1199744/app-debug.apk.zip)

CloudyPadmal:
Thanks a lot @viveksb007 for the review! I was wondering what if we put a slider here instead of an edit text? :thinking:

viveksb007:
I think EditText is better. Slider actually is difficult to adjust to specific values and we don't need it continuously. Its one time input to start experiment.

CloudyPadmal:
Yea that's a very good point! :+1:

Issue #386, opened 2017-07-29
log data on selecting sensor from sensor list


viveksb007:
**Actual Behaviour**

Currently DataSensorLogger Activity only scans and shows sensors as a list.

**Expected Behaviour**

It should log data on selecting sensor from sensor list displayed after scanning. Parent Issue #363 

**Steps to reproduce it**

Application Fragment -> DataSensorLogger Activity

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

Yes

PR #427, opened 2017-08-04
added realm dependency and created data model for sensor MPU6050

viveksb007:
Partially Fixes issue #386 

Changes: 
- Added [Realm](https://realm.io/docs/java/latest/) dependency
- Added data model for sensor MPU6050


app/src/main/java/org/fossasia/pslab/MyApplication.java
```diff
@@ -0,0 +1,18 @@
...
+
+import android.app.Application;
+
+import io.realm.Realm;
+
+/**
+ * Created by viveksb007 on 4/8/17.
+ */
+
+public class MyApplication extends Application {
```
CloudyPadmal: Shall we use `PSLabApplication` instead of `MyApplication`? :)

viveksb007: @CloudyPadmal Sounds good. Made changes.

CloudyPadmal: Awesome! :100:

Issue #434, opened 2017-08-06
Add WirelessSensor Activity


asitava1998:
**Actual Behaviour**
Currently, there is no activity for Wireless sensors.

**Expected Behaviour**
A dedicated activity for Wireless Sensors.

**Steps to reproduce it**
Main Menu -> Applications -> Wireless Sensors

**LogCat for the issue**
None

**Screenshots of the issue**
None

**Would you like to work on the issue?**
Yes

PR #435, opened 2017-08-06
Creates activity for Wireless Sensors

asitava1998:
Fixes issue #434 

Changes:
* Creates an activity for Wireless Sensors.
* Creates corresponding layout file.

Detailed implementation can be discussed in another issue.

Screenshots for the change: 
None

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/activity/WirelessSensorActivity.java
+++ b/app/src/main/java/org/fossasia/pslab/activity/WirelessSensorActivity.java
@@ -1,0 +1,17 @@
+package org.fossasia.pslab.activity;
+
+import android.os.Bundle;
+import android.support.annotation.Nullable;
+import android.support.v7.app.AppCompatActivity;
+
+import org.fossasia.pslab.R;
+
+
+public class WirelessSensorActivity extends AppCompatActivity {
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_wireless_sensor);
+    }
+}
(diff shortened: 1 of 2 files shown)


viveksb007:
@asitava1998 Pocket Version doesn't support communication with wireless sensors.
@jithinbp what should we do here?

asitava1998:
@viveksb007 I didn't know that . The NRF24L01 module cannot be used with PSLab to be used with Wireless sensors?

jithinbp:
@viveksb007 @asitava1998  hey, it was not intended to be used with wireless sensors, and there's no space to solder the NRF24 module ( no vertical room. The footprint is there. )
I've added the NRF's libraries to the firmware repository (pslab-firmware), but the actual devices do not have this firmware. If you wish, you can solder the nrf modules (but you won't be able to replace the top panel due to space constraints.) , and upgrade the firmware by following the instructions on the repository. However it's probably best if you focus on other functions such as the experiment designer at this point. I have also not manufactured the wireless sensor nodes to send, so it won't be the best use of your time to handle wireless functions at this point.

asitava1998:
@viveksb007  Then it is better to modify this PR and remove this option of Wireless Sensors as it not serving a purpose as of now. What do you think?
But with 5 tiles, the UI will look odd.

viveksb007:
I think we should let Application Fragment be as it is. We would work on it later when things at hand would be completed.
Closing this PR for now.

jithinbp:
okay, cool. 5 tiles is not an issue coz this is under active development, and i'm sure there are a lot of new ideas that can be added .

CloudyPadmal:
We can add `Science Journal` as the sixth tile just in case :D

Issue #437, opened 2017-08-06
Remove Wireless sensor tile from Application tile


viveksb007:
**Actual Behaviour**

Wireless Sensor tile isn't required as Pocket Version doesn't support it. 

**Expected Behaviour**

Remove Wireless Sensor tile

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Would you like to work on the issue?**

@asitava1998

PR #447, opened 2017-08-06
Removes Wireless Sensor from Application

asitava1998:
Fixes issue #437 

Changes: 
* Removed Wireless Sensors from Applications tile as the feature is not yet developed and is redundant as of now.
* A minor bug fix of Sensor QuickView crashing

Screenshots for the change:


CloudyPadmal:
We can later integrate some other option here right? :)

PR #512, opened 2017-08-14
Fixes #511 Fixed buggy behaviour of control main fragment

### LINKED ISSUE #511: editTextValue changed on Set button click in Control Main
**Actual Behaviour**

After seekbar is changed from default value, when `Set` button is clicked it changes editText value at each click.

**Expected Behaviour**

On clicking `Set` button only pslab device should be configured, EditText value shouldn't change. 

**Steps to reproduce it**

Control Main -> Change seekbar of W1 -> click on `Set` button again and again, value of edit text would change on each click. 

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

Yes

viveksb007:
Fixes issue #511 

Changes:
- Prevented unwanted trigger in EditText value by SeekBar.
- Modified map() function to support double data types.

apk for testing : 
[app-debug.apk.tar.gz](https://github.com/fossasia/pslab-android/files/1221796/app-debug.apk.tar.gz)

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/adapters/ControlMainAdapter.java
+++ b/app/src/main/java/org/fossasia/pslab/adapters/ControlMainAdapter.java
@@ -26,6 +26,7 @@ public class ControlMainAdapter extends RecyclerView.Adapter<ControlMainAdapter.
 
     private String[] mDataset;
     private ScienceLab scienceLab = ScienceLabCommon.scienceLab;
+    private boolean manualSeekBarChange = false;
 
     static class ViewHolder extends RecyclerView.ViewHolder {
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/pslab/adapters/ControlMainAdapter.java
```diff
@@ -550,14 +597,16 @@ else if (value > 5000)
                     }
                 });
 
-                seekBarControlMain.setMax(5000);
```
viveksb007: Because previously range of seekbar was 0-5000 but allowed values were 10-5000. So they were not accurately mapped. 0-4990 -> 10-5000 hope this clears the doubt.

CloudyPadmal: Yea it does! Thanks @viveksb007 :+1:

Issue #483, opened 2017-08-11
Implement Ramp Generator Experiment under Op-Amp circuits


asitava1998:
**Actual Behaviour**
Ramp Generator experiment under Op-Amp circuits is not yet implemented

**Expected Behaviour**
Ramp Generator experiment under Op-Amp circuits needs to be implemented

**Steps to reproduce it**
1. Open app
2. Go to `Saved Experiments`
3. Select `Electronics Experiments`
4. Select `OpAmp circuits`
5. Select `Ramp Generator`

**LogCat for the issue**
> None

**Screenshots of the issue**
> None

**Would you like to work on the issue?**
Yes

PR #517, opened 2017-08-15
Ramp Generator Experiment

asitava1998:
Fixes issue #483 

Changes: 
* Creates the Ramp Generator Experiment
Note: The user doesn't need to set the frequency of SQR1. so the frequency of SQR1 is set to a constant value of 1000 Hz .

Screenshots for the change: 

> The images above are not of actual experiment.
[app.zip](https://github.com/fossasia/pslab-android/files/1226822/app.zip)


mariobehling:
Please resolve conflicts.

Issue #529, opened 2017-08-17
Implement Summing Junction Experiment


asitava1998:
**Actual Behaviour**
Summing Junction Experiment under Op-Amp circuits is not yet implemented

**Expected Behaviour**
Summing Junction Experiment under Op-Amp circuits needs to be implemented

**Steps to reproduce it**
1. Open app
2. Go to `Saved Experiments`
3. Select `Electronics Experiments`
4. Select `OpAmp circuits`
5. Select `Summing Junction`

**LogCat for the issue**
> None

**Screenshots of the issue**
> None

**Would you like to work on the issue?**
Yes

PR #533, opened 2017-08-18
Creates Summing Junction Experiment

asitava1998:
Fixes issue #529 

Changes: 
* Creates the summing junction experiment

Screenshots for the change: 
[app.zip](https://github.com/fossasia/pslab-android/files/1237884/app.zip)


app/src/main/java/org/fossasia/pslab/experimentsetup/SummingJunctionExperiment.java
```diff
@@ -0,0 +1,321 @@
...
+import java.util.List;
+
+/**
+ * Created by asitava on 9/8/17.
+ */
+
+public class SummingJunctionExperiment extends Fragment{
+
+    private static final String ERROR_MESSAGE = "Invalid Value";
+    private static final String MINIMUM_VALUE_FREQUENCY = "Frequency should be more than 10Hz";
```
jithinbp: LInes such as these are being repeated across experiments. Can't we put them in a single file and reference them?
And instead of writing the whole string, we can just have simple arrays for each I/O and generate it. e.g. ['WG1','Hz',10,5000]

asitava1998: I am not aware whether this is possible. Maybe @viveksb007 , @akarshan96 and @CloudyPadmal  can tell better if this can be implemented

CloudyPadmal: Hi! Yes @jithinbp these validation lines are available almost in every experiment. We can create a java file to include all the error messages and parameters as you suggested! :)

viveksb007: Yeah all these `static` strings can be moved to one file and can be accessed from there, @asitava1998 its something similar to what we did with commandsProto.java

app/src/main/res/values/strings.xml
```diff
@@ -383,6 +383,18 @@
...
+    <string-array name="wave_generator">
+        <item>W1</item>
+        <item>W2</item>
+    </string-array>
+
+    <string-array name="voltage_source">
+        <item>PV1</item>
+        <item>PV2</item>
+        <item>PV3</item>
+        <item>PCS</item>
```
viveksb007: `PCS` isn't a voltage source.

app/src/main/res/layout/summing_junction_setup.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
viveksb007: This layout file is similar to other experiments like diode, zener, etc. So we can use other experiment layout and just add a comment that reusing **which** experiment setup and re-use layout file.

app/src/main/res/values/strings.xml
asitava1998: Yes. I know it is a current source. I wanted to use them in the same spinner. I just kept the name as `voltage_source`. I can change the name. Suggest an alternative?

app/src/main/java/org/fossasia/pslab/experimentsetup/SummingJunctionExperiment.java
asitava1998: I can handle them in a separate issue and implement them in all the files.

app/src/main/res/values/strings.xml
viveksb007: `source_pin` maybe?
This would also work, not much issue here.

asitava1998: Sure 👍

asitava1998:
@viveksb007 I reused the `diode_setup.xml` file and changed the spinner name. Changes done.

app/src/main/java/org/fossasia/pslab/experimentsetup/SummingJunctionExperiment.java
jithinbp: cool, thanks. should save time spent on developing further experiments

PR #550, opened 2017-08-20
Fixes #549 : PerformExperiments adapter switch cases

### LINKED ISSUE #549: PerformExperimentAdapter contains lot of if clauses
**Actual Behaviour**

`PerformExperimentAdapter` contains lot of if clauses inside `public Fragment getItem(int position)` method to configure the relevant fragment. As in this case every if clause is evaluated if the last fragment is the one to be processed, it is not a good programming practice to use if clauses like that

**Expected Behaviour**

These if clauses should be replaced with switch case statements to improve performance.

**Steps to reproduce it**

Refer the [java](https://github.com/fossasia/pslab-android/blob/development/app/src/main/java/org/fossasia/pslab/adapters/PerformExperimentAdapter.java#L49) file lines

**LogCat for the issue**

> Not available

**Screenshots of the issue**

> Not available

Yes

CloudyPadmal:
Fixes issue #549 

Changes: 
- Added switch cases to if statements
- Since values needed to be constant, couldn't use string resources

APK for testing:
[app-debug.apk.zip](https://github.com/fossasia/pslab-android/files/1236842/app-debug.apk.zip)

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/adapters/PerformExperimentAdapter.java
+++ b/app/src/main/java/org/fossasia/pslab/adapters/PerformExperimentAdapter.java
@@ -49,130 +49,123 @@ public class PerformExperimentAdapter extends FragmentPagerAdapter {
     public Fragment getItem(int position) {
         switch (position) {
             case 0:
-                if (experimentTitle.equals(context.getResources().getString(R.string.diode_iv)))
-                    return ExperimentDocFragment.newInstance("D_diodeIV.html");
-                if (experimentTitle.equals(context.getResources().getString(R.string.zener_iv)))
-                    return ExperimentDocFragment.newInstance("D_ZenerIV.html");
-                if (experimentTitle.equals(context.getResources().getString(R.string.half_wave_rectifier)))
-                    return ExperimentDocFragment.newInstance("L_halfWave.html");
-                if (experimentTitle.equals(context.getResources().getString(R.string.transistor_cb)))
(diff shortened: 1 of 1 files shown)


viveksb007:
@CloudyPadmal I don't think this improves much of performance and we are losing flexibility too. Any change in experiment title won't be reflected here. 
What do you think?

CloudyPadmal:
Hi @viveksb007 ! It will improve performance as there are a lot of saved experiments. Please [refer](https://stackoverflow.com/questions/4334339/disadvantages-to-using-lots-of-if-statements) to this article. There will be a lag when there are many experiments are implemented and user clicked on one of the last experiments if `if` statements are used.

viveksb007:
From what I understand. A lot of comparisons are made in case below, and all conditions are mutually exclusive
```
if(condition)
// do something
if(condition)
// do  something
if(condition)
// do  something
```
But in our case its,
```
if(condition)
return
if(condition)
return
if(condition)
return
```
So performance is same. Same comparisons would be made in our current implementation and the switch case you propose.

CloudyPadmal:
Switch case uses hash tables which makes it faster in accessing elements where mutually exclusive if statements are dependable on what item user clicks. [Reference](https://stackoverflow.com/questions/22110707/how-is-string-in-switch-statement-more-efficient-than-corresponding-if-else-stat)

viveksb007:
@CloudyPadmal Thanks for the reference that cleared the doubt.

CloudyPadmal:
Thanks all for the review! :+1:

Issue #569, opened 2017-08-22
same set of strings are used across all experiments


viveksb007:
**Actual Behaviour**

Same set of strings are used across all experiments, which are declared separately in all experiment files. 

**Expected Behaviour**

All strings can be defined into one file and can be referred from there.

**Steps to reproduce it**

NA

**LogCat for the issue**

NA

**Screenshots of the issue**

NA

**Would you like to work on the issue?**

@asitava1998

PR #593, opened 2017-08-27
Creates a Common Error String Resource File

asitava1998:
Fixes issue #569 

Changes: 
* Implements same set of strings across all experiment files
* Creates a common file for the same

Screenshots for the change: 
None

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/experimentsetup/ExperimentErrorStrings.java
+++ b/app/src/main/java/org/fossasia/pslab/experimentsetup/ExperimentErrorStrings.java
@@ -1,0 +1,27 @@
+package org.fossasia.pslab.experimentsetup;
+
+/**
+ * Created by asitava on 27/8/17.
+ */
+
+public class ExperimentErrorStrings {
+    static String ERROR_MESSAGE = "Invalid Value";
+    static String INVALID_VALUE = "Voltage value too low";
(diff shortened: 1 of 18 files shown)


CloudyPadmal:
Hi @asitava1998 ! It's a good fix! What if we use a static class for strings? :)

asitava1998:
@CloudyPadmal  We can do that too. Should I declare it as static then?

CloudyPadmal:
Hi @asitava1998 ! If it's not much of a trouble! :)

app/src/main/java/org/fossasia/pslab/experimentsetup/ExperimentErrorStrings.java
```diff
@@ -0,0 +1,27 @@
+package org.fossasia.pslab.experimentsetup;
+
+/**
+ * Created by asitava on 27/8/17.
+ */
+
+public class ExperimentErrorStrings {
+    String ERROR_MESSAGE = "Invalid Value";
```
viveksb007: You can create a `static` class and define all `strings` as `static`. Then you won't need to create `ExperimentErrorStrings` object. Simply refer to `ExperimentErrorStrings.anystring`

asitava1998:
@CloudyPadmal @viveksb007 I have made the changes. Declared the strings static.

PR #628, opened 2017-09-15
fixes #627 add double tap to exit

### LINKED ISSUE #627: On a single back press I came out of the app
**Actual Behaviour**

When we are on the home fragment and press the back button,we exit the app.

**Expected Behaviour**

There should be a toast message like press back again to close.

**Steps to reproduce it**

Open the app and press the back button

Yes

nikit19:
Fixes issue #627 

Changes: added double tap to exit

Screenshots for the change: 
ezgif com-video-to-gif 2

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/activity/MainActivity.java
+++ b/app/src/main/java/org/fossasia/pslab/activity/MainActivity.java
@@ -24,6 +24,7 @@ import android.view.MenuItem;
 import android.view.View;
 import android.widget.ImageView;
 import android.widget.TextView;
+import android.widget.Toast;
 
 
 import org.fossasia.pslab.communication.CommunicationHandler;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/fossasia/pslab/activity/MainActivity.java
```diff
@@ -252,6 +255,7 @@ private void loadNavHeader() {
 
     @Override
     public void onBackPressed() {
+              Fragment fragment = getSupportFragmentManager().findFragmentById(R.id.frame);
```
viveksb007: reformat the code.

nikit19: @viveksb007 is it fine now?

PR #653, opened 2017-09-30
Fix #652 landscape rotation works now

### LINKED ISSUE #652: OscilloscopeActivity doesn't rotate on screen rotation
**Actual Behaviour**

OscilloscopeActivity remains fixed on screen rotation

**Expected Behaviour**

It should remain in landscape mode but allow the user to rotatate the screen if the user desires to do so

**Steps to reproduce it**

Open the app. Select applications and click on Oscilloscope

Yes

nikit19:
Fixes issue #652 

Changes: landscape rotation works now

Screenshots for the change: 
ezgif com-video-to-gif 11

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -46,8 +46,8 @@
             android:configChanges="orientation|screenSize" />
         <activity
             android:name=".activity.OscilloscopeActivity"
-            android:configChanges="orientation|screenSize|keyboardHidden"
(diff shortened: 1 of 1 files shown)


viveksb007:
Please add an `apk` for testing with device.

nikit19:
@viveksb007 
here is the apk https://drive.google.com/file/d/0B0Nwrte8-rWVVU9oUkNoWTN6T1E/view?usp=sharing

Issue #673, opened 2017-10-25
Replace hardcore strings with string resource,wherever it is not present


akashagra:
Several times hardcore strings are used,which is not a good practice ,hence replace them with string resource.

divyabaid16:
Can you please elaborate this issue?

arya-ayush:
I will fix it .

asitava1998:
Fixed in PR #674

PR #674, opened 2017-10-25
Replaced Hardcore Strings with String Resource

akashagra:
Replaced all hardcore strings with string resource and fixed an error that was caused on api 17 or lower.
#673


akashagra:
@mariobehling can you please suggest why codacy is not able to analyse my pull request?

akashagra:
This branch is still not merged .Kindly look into this @the-dagger

PR #680, opened 2017-12-05
Redundant play button removed

Anshpuri:
Fixes issue #[679]

Changes: [Redundant play button removed]

Screenshots for the change:


viveksb007:
@Anshpuri There are many commits in the PR that are not yours. I think you messed up during squashing commits. Please try squashing commits and send the PR with only single commit that solves the issue.
Cheers.

Anshpuri:
@viveksb007 sir please advice how do I do that

PR #695, opened 2018-01-12
Fix #694 cursor is positioned at the end of editText

### LINKED ISSUE #694: Cursor position should be at the end of the editText
**Actual Behaviour**

Currently when we open the ControlFragmentAdvanced,and select an editText the cursor is positioned at the beginning of the  editText and the user has to the move the cursor to the end and then change the value.

**Expected Behaviour**

Cursor should be present at the end of the editText

**Steps to reproduce it**

Open the app.
Select applications.
Open Control.
Select Advanced fragment
Click on any editText

**Screenshots of the issue**

screenshot_20180112-181403

Yes

nikit19:
Fixes issue #694 

Changes:Cursor is positioned at the end of editText

Screenshots for the change: 
ezgif com-video-to-gif 30

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/others/EditTextWidget.java
+++ b/app/src/main/java/org/fossasia/pslab/others/EditTextWidget.java
@@ -75,8 +75,10 @@ public class EditTextWidget extends LinearLayout{
                     data = data > maxima ? maxima : data;
                     data = data < minima ? minima : data;
                     editText.setText(String.valueOf(data));
+                    editText.setSelection(String.valueOf(data).length());
                 } catch (Exception e) {
                     editText.setText("0");
+                    editText.setSelection(1);
                 }
             }
         });
@@ -90,8 +92,10 @@ public class EditTextWidget extends LinearLayout{
                     data = data > maxima ? maxima : data;
                     data = data < minima ? minima : data;
                     editText.setText(String.valueOf(data));
+                    editText.setSelection(String.valueOf(data).length());
                 } catch (Exception e) {
                     editText.setText("0");
+                    editText.setSelection(1);
                 }
             }
         });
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/pslab/others/EditTextWidget.java
```diff
@@ -90,8 +92,10 @@ public void onClick(View v) {
                     data = data > maxima ? maxima : data;
                     data = data < minima ? minima : data;
                     editText.setText(String.valueOf(data));
+                    editText.setSelection(String.valueOf(data).length());
                 } catch (Exception e) {
                     editText.setText("0");
+                    editText.setSelection("0".length());
```
CloudyPadmal: Could this be simply put as 1?

app/src/main/java/org/fossasia/pslab/others/EditTextWidget.java
```diff
@@ -122,5 +126,6 @@ public String getText() {
 
     public void setText(String text) {
```
CloudyPadmal: I wonder what the use of this function! :thinking: in the onclick implementation it's the same lines is it?

nikit19: Actually the onClick implementation is for the button clicks. When the value gets incremented or decremented the cursor will also remain at the end.
While this setText method is for EditTextWidget so when this dialog is opened,the cursor will be seen at the end

app/src/main/java/org/fossasia/pslab/others/EditTextWidget.java
nikit19: Sure :+1:

CloudyPadmal:
Could you please check why the Travis build is failing? PR looks great! 👍

nikit19:
@CloudyPadmal travis build has passed now

Issue #704, opened 2018-01-24
Display a temporary toast message when a non-working button is clicked


abhinavraj23:
**Actual Behaviour**

When a non working button such as in the buttons of physics experiment(saved experiments) is clicked nothing happens. 

**Expected Behaviour**

There should be a toast message displaying "Will be available soon".

**Steps to reproduce it**

Display toast message while the button is clicked.

**Would you like to work on the issue?**
Yes.

Parag09:
i would like to work on it

PR #705, opened 2018-01-26
Fixed issue #704 Added a toast message for non working buttons

abhinavraj23:
Fixes issue #704

Changes: Added toast message for non working buttons (which were in the saved experiments section) which displays "Will be available soon"

Screenshots for the change:
pslabscreenshot


app/src/main/java/org/fossasia/pslab/fragment/SavedExperiments.java
```diff
@@ -157,6 +158,7 @@ public void onClick(TreeNode node, Object value) {
                     @Override
                     public void onClick(TreeNode node, Object value) {
                         // Open BJT CE Backup experiment
+                        Toast.makeText(getActivity(), "Will be available soon", Toast.LENGTH_LONG).show();
```
viveksb007: Make Toast.LENGTH_SHORT in all the Toasts used.

abhinavraj23: Will surely do that @viveksb007

Issue #736, opened 2018-02-25
Bug in the Control Advanced fragment


Avjeet:
**Actual Behaviour**

 The value of all the  edittext in arbitrary waveform generator  is being set to zero on pressing the set button.

**Expected Behaviour**

It should not set the value to zero.

**Steps to reproduce it**

activity->control activity->control advanced fragment

**Screenshots of the issue**
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/20573611/36640424-5df6b814-1a44-11e8-8fca-16a2daaa0511.gif)

**Would you like to work on the issue?**
Yes

Avjeet:
@CloudyPadmal @viveksb007 @akarshan96 It is due to the phase editext is throwing number format exception as it is expecting a value in integer and the editext has float,mentors plz guide what value is required by the phase (float or integer ) so that I can make the changes accordingly.

CloudyPadmal:
Resolved in #748

PR #748, opened 2018-03-04
Bug resolved in control advanced fragment

Avjeet:
First steps of issue #736 

Changes: 
             changed the datatype of the phase which is causing Number Format exception.


Avjeet:
@CloudyPadmal This phase variable is unused so plz guide how this "phase" is used so as to resolve this issue properly.

CloudyPadmal:
@Avjeet the phase is used to offset two waveforms and it will be used in wave generator implementation.

akarshan96:
@Avjeet is this PR work in progress or ready for review as it mentions First steps of issue #736

Avjeet:
@akarshan96 yes , actually i have a problem in understanding where this value of phase is used in the communication module, as in how should I store this value of phase in the science lab class , Should I make a new variable and store the value.

Issue #774, opened 2018-03-30
Add option menu in App bar


ThisIsNSH:
**Actual Behaviour**
Settings fragment & few other options are currently present in the navigation drawer.

**Expected Behaviour**
All these fragments which can be piled up should be put in the menu of app bar.

**Steps to reproduce it**
Create menu.xml and implement menu in app bar.

**Screenshots of the issue**
![screenshot_2018-03-30-14-28-06-009_org fossasia pslab](https://user-images.githubusercontent.com/32385519/38132553-25547542-3429-11e8-9e23-2b0940330831.png)

**Suggested mock**
![147938263429358](https://user-images.githubusercontent.com/32385519/38132345-5a2d0cbc-3428-11e8-900a-d9d4e095a2d7.png)

**Would you like to work on the issue?**
Yes

ThisIsNSH:
@mariobehling Yes we can do it that way. But we can also provide different options for each fragment in the menu of the app bar.

ThisIsNSH:
Is this a good idea?

PR #775, opened 2018-03-30
Setting option added on appbar

ThisIsNSH:
Fixes issue #774 

Changes: The settings option was removed from the drawer and added to the appbar. This will allow user to access not only settings but many other options from any fragments (which will be added later). Back button functionality also works and the menu option become invisible when on settings fragment.

Screenshots for the change: 

APK for testing: [Compress the *.apk file into a rar or zip file and upload it here]
[app-debug.zip](https://github.com/fossasia/pslab-android/files/1863378/app-debug.zip)


ThisIsNSH:
@CloudyPadmal @akarshan96 @viveksb007 please review. This menu option will be of great help to add more shortcuts in future.

Issue #776, opened 2018-03-30
Add Different language support


ThisIsNSH:
**Actual Behaviour**
Currently the app only has English language. Chinese is one of the most spoken language in the world. Not only Chinese but other languages should be added to increase the usability of the app. 

**Expected Behavior**
Users should be able to view app in other languages too.

**Steps to reproduce it**
Start with creating string.xml in chinese and then move on to other languages

**Would you like to work on the issue?**
yes

CloudyPadmal:
We can consider about language support in a latter phase

PR #777, opened 2018-04-01
values-zh/string.xml created

ThisIsNSH:
First Step for issue #776 

Changes: Converted string.xml into chinese.

Screenshots for the change: None 

APK for testing: None

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'realm-android'
 
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
-    buildToolsVersion rootProject.ext.buildToolsVersion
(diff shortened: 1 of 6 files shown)


ThisIsNSH:
@CloudyPadmal @akarshan96 @viveksb007 please review the changes in string.xml. 
I have chosen Chinese as it is one of the most spoken language. Further I will work on other languages to make the app localized.
Please suggest some languages if the idea of localization is ok.

ThisIsNSH:
I used google translation @mariobehling. Suggest me something..

CloudyPadmal:
@ThisIsNSH accuracy of Google translation would be questionable. A translation better done by a native speaker!

CloudyPadmal:
@ThisIsNSH yes that's nice! :100:

Issue #782, opened 2018-04-12
Take out all "Closing" dialogues


mariobehling:
Avoid dialogues like "Do you want to close this" if the user uses the back button. Delete these dialogues and perform the action without dialogue.

Avjeet:
I want to work on this.

PR #787, opened 2018-04-12
Removed the exit dialog box from every activity

Avjeet:
Fixes issue #782 
Changes:Remove the dialog box code from every activity  in OnBackPressed() .

APK for testing: 
[app-debug.zip](https://github.com/fossasia/pslab-android/files/1903891/app-debug.zip)

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/activity/ControlActivity.java
+++ b/app/src/main/java/org/fossasia/pslab/activity/ControlActivity.java
@@ -73,20 +73,7 @@ public class ControlActivity extends AppCompatActivity {
 
     @Override
     public void onBackPressed() {
-        new AlertDialog.Builder(this)
-                .setIcon(android.R.drawable.ic_dialog_alert)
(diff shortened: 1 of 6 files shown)


harsh-2711:
I think there's no need for removing this method from every activity as confirmation is always better while exiting or else there might arise a situation where by mistake also(back press), whole experiment might get affected.

Avjeet:
@harsh-2711 The dialog should only be once when the user need to exit the app ,asking for confirmation every time the user closes one activity could lead to bad user experience and also the state of the activity can be saved before closing it.

harsh-2711:
@Avjeet  What are you telling is correct but saving each and every state where the user visits is a bit difficult and memory consuming. Else it is better to ask for confirmation and then close the activity.

Avjeet:
Only some activities like control activity requires the setting to be saved else can be created again and having dialog box asking everytime disrupts the flow of the app.

PR #870, opened 2018-05-12
fix: application can connect with the device

### LINKED ISSUE #834: Application sometimes doesn't detect the device
**Actual Behaviour**

Sometimes we have to unplug and plug back the device for the app to detect

**Expected Behaviour**

Whenever the device is attached, application should detect it.

**Steps to reproduce it**

1. Open app
2. Connect the device (This happens rarely but that is a big issue)

**LogCat for the issue**

> Not available

**Screenshots of the issue**

> Not available

Yes

CloudyPadmal:
Fixes #834 

Changes: 
- Added a menu item to connect the PSLab without manually plugging off and on.
- Update connection status on the menu bar by changing icons
- Added device initialization
- Added device ID to navigation menu

Screenshots for the change: 
recording

APK for testing:
[app-debug.apk.zip](https://github.com/fossasia/pslab-android/files/1997841/app-debug.apk.zip)


app/src/main/java/org/fossasia/pslab/activity/MainActivity.java
```diff
@@ -76,6 +83,10 @@
     private Handler mHandler;
     private ScienceLabCommon mScienceLabCommon;
 
+    public boolean PSLabisConnected = false;
+
+    InitializationVariable initialisationStatus;
```
Vikum94: Can this be private?

CloudyPadmal: Hi @Vikum94 ! It is used in another fragment in the same package. So I left it in the package scope!

CloudyPadmal:
Thank you @akarshan96 :+1:

mariobehling:
Please resolve conflicts.

PR #886, opened 2018-05-18
fix: gradle versions in config files

### LINKED ISSUE #883: Gradle file version issues
**Actual Behaviour**

There are few libraries which were added recently with their versions hardcoded.

**Expected Behaviour**

Similar to earlier libraries, version should be stated in config files

**Steps to reproduce it**

Check the build.gradle file

**LogCat for the issue**

```
    implementation "com.jakewharton:butterknife:$rootProject.butterKnifeVersion"
    implementation 'com.android.support:support-v4:27.1.1'
    annotationProcessor "com.jakewharton:butterknife-compiler:$rootProject.butterKnifeVersion"
    implementation 'com.github.medyo:android-about-page:1.2.4'
    implementation 'com.github.tiagohm.MarkdownView:library:0.19.0'
    debugImplementation 'com.squareup.leakcanary:leakcanary-android:1.5.4'
    releaseImplementation 'com.squareup.leakcanary:leakcanary-android-no-op:1.5.4'
    testImplementation 'com.squareup.leakcanary:leakcanary-android-no-op:1.5.4'

    compile 'com.github.mirrajabi:search-dialog:1.1'
```

**Screenshots of the issue**

> Not available

Yes

CloudyPadmal:
Fixes #883 

Changes:
- Added versions to config file and called from there

Screenshots for the change: 

> Not available

APK for testing: 
[app-debug.apk.zip](https://github.com/fossasia/pslab-android/files/2017629/app-debug.apk.zip)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -44,15 +44,14 @@ dependencies {
     implementation "com.github.devlight.navigationtabstrip:navigationtabstrip:$rootProject.navTabStripVersion"
     implementation "com.afollestad.material-dialogs:core:$rootProject.materialDialogsVersion"
     implementation "com.jakewharton:butterknife:$rootProject.butterKnifeVersion"
-    implementation 'com.android.support:support-v4:27.1.1'
+    implementation "com.android.support:support-v4:$rootProject.supportLibraryVersion"
     annotationProcessor "com.jakewharton:butterknife-compiler:$rootProject.butterKnifeVersion"
-    implementation 'com.github.medyo:android-about-page:1.2.4'
-    implementation 'com.github.tiagohm.MarkdownView:library:0.19.0'
-    debugImplementation 'com.squareup.leakcanary:leakcanary-android:1.5.4'
-    releaseImplementation 'com.squareup.leakcanary:leakcanary-android-no-op:1.5.4'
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -44,15 +44,14 @@ dependencies {
...
-    implementation 'com.github.medyo:android-about-page:1.2.4'
-    implementation 'com.github.tiagohm.MarkdownView:library:0.19.0'
-    debugImplementation 'com.squareup.leakcanary:leakcanary-android:1.5.4'
-    releaseImplementation 'com.squareup.leakcanary:leakcanary-android-no-op:1.5.4'
-    testImplementation 'com.squareup.leakcanary:leakcanary-android-no-op:1.5.4'
-
-    compile 'com.github.mirrajabi:search-dialog:1.1'
+    implementation "com.github.medyo:android-about-page:$rootProject.androidAboutPageVersion"
+    implementation "com.github.tiagohm.MarkdownView:library:$rootProject.markDownViewVersion"
+    implementation "com.github.mirrajabi:search-dialog:$rootProject.searchDialogVersion"
```
harsh-2711: I think this will use Search Dialog Version 1.2.1 and so it will create issues while again implementing search bar as it is the latest version with some changes that needs to be defined for the search bar to work. So, I suggest you to keep hardcoded version

PR #896, opened 2018-05-21
Added colors for different channels in Logic Analyzer instrument

### LINKED ISSUE #895: Missing different colors for different channels in Logic Analyzer Instrument
**Actual Behaviour**

The color of every channel remains same and also the legends have same colors.

**Expected Behaviour**

There should be different colors for different channels so that they can be differentiated

**Steps to reproduce it**

Go to Logic Analyzer instrument and select more than one channel mode and then press analyze button

**LogCat for the issue**

Not applicable

Yes

harsh-2711:
Fixes #895 

Changes: Added different colors for different channel in every mode along with adding colors in legend

Screenshots for the change: 
screenshot_20180522-015828
screenshot_20180522-015841

APK for testing: 
[app-debug.zip](https://github.com/fossasia/pslab-android/files/2024002/app-debug.zip)

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
+++ b/app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
@@ -98,12 +98,6 @@ public class LALogicLinesFragment extends Fragment {
                                 }
                             });
                         }
-                        // TODO : Perform Logic Analyzer related functions
-                        // Colors for Data sets
-                        // - CH1 - (Color.BLUE);
-                        // - CH2 - (Color.GREEN);
-                        // - CH3 - (Color.RED);
-                        // - CH4 - (Color.YELLOW);
                     } else {
                         if (!String.valueOf(ledImageView.getTag()).equals("red")) {
                             getActivity().runOnUiThread(new Runnable() {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/fossasia/pslab/fragment/LALogicLinesFragment.java
```diff
@@ -181,7 +181,24 @@ private void updateLogicLines() {
             }
             LineDataSet lineDataSet = new LineDataSet(tempInput, channelNames.get(j));
             lineDataSet.setCircleRadius(1);
-            lineDataSet.setColor(Color.GREEN);
+            switch (j) {
+                case 0:
+                    lineDataSet.setColor(Color.BLUE);
```
CloudyPadmal: This color is very hard to see in a black background. Always think of a user's point of view.

yatri1609:
@harsh-2711 It would be nice if you could increase the thickness of lines a bit. Also Red color is something which is not much going with the black background, try using bright colors ( maybe white will work well ).

harsh-2711:
@yatri1609  I can't use white color as it matches with legend color. But I will try to add more bright colors as @CloudyPadmal  said and will surely increase the width of the lines

harsh-2711:
screenshot_20180522-153814
screenshot_20180522-153819
Are these fine?

CloudyPadmal:
These colors look good! :+1:

PR #905, opened 2018-05-22
feat: added interactive pin layout diagram

### LINKED ISSUE #875: Create a pin layout image
**Actual Behaviour**

PSLab devices come in two versions and another one is being finalized.
- CSPark
- PSLab V4
- PSLab V5

They have different pin placements and sizes. A guide is lacking for user to identify which pin does what and where is that pin.

**Expected Behaviour**

The opening screen is not having any functionality at the current app. This page can be used to layout a pin diagram with actual images of PSLabs where user can click on each pin so a dialog box appears describing what the pin is and it's capabilities.

**Steps to reproduce it**

1. Open app and see the home page

**LogCat for the issue**

> Not available

**Screenshots of the issue**

> Not available

Yes

CloudyPadmal:
Fixes #875 

Changes:
- Added an interactive PSLab layout diagram for users to navigate by zooming and clicking on each pin separately.

Screenshots for the change: 
device-2018-05-22-144743

This is a low-res screen-cast;
[PinLayouts.mp4.zip](https://github.com/fossasia/pslab-android/files/2025854/PinLayouts.mp4.zip)

APK for testing: 
[app-debug.apk.zip](https://github.com/fossasia/pslab-android/files/2026728/app-debug.apk.zip)


harsh-2711:
I found out some bugs which needs to be fixed

1. The description of some pins gets cut out as shown
screenshot_20180522-150526

2. I think you should remove cancel button and provide only description as clicking on screen other than dialog will close it and so cancel button isn't necessary. In this way the description won't get cut.

3. Also add some constraints to zoom in and zoom out as if we do more zoom in than app crashes and if zoom out is done more the layout disappears

CloudyPadmal:
@harsh-2711 please download and check the latest apk. You must have downloaded the earlier version where I fixed the font size.

harsh-2711:
@CloudyPadmal  I downloaded the apk from APK for testing section. If there is any latest, can you  please provide it?

CloudyPadmal:
APK was updated 10 mins after the PR was made.

cweitat:
is it possible to restrict the max zoom out size? currently user is able to zoom out till very small.

CloudyPadmal:
Yes @cweitat ! Let me give it a try! :)

PR #884, opened 2018-05-18
feat: Added card view for instruments section

harsh-2711:
Fixes #877 

Changes: Added Card View in instruments section with 1 column in portrait mode and 2 columns in landscape mode. Also great images are added as background.

Screenshots for the change: 
**Portrait Mode**

screenshot_20180518-044330

**Landscape Mode**

screenshot_20180518-102344

APK for testing: 
[app-debug.zip](https://github.com/fossasia/pslab-android/files/2015627/app-debug.zip)

PR #906, opened 2018-05-23
feat: created multimeter activity

### LINKED ISSUE #885: Create the multimeter module
**Actual Behaviour**

Currently the multimeter is inside the control unit section.

**Expected Behaviour**

It has to be made a separate module itself.

**Steps to reproduce it**

Create a multimeter fragment inside the instrument fragment and implement all it's functionality.

**LogCat for the issue**

Not applicable

Yes

abhinavraj23:
Fixes #885 partially

Changes: Created the multimeter activity but the layout is yet to be changed. 

APK for testing: The apk will only make sense if the pr #884 is merged as there is not different multimeter section in the previous version of app so there is no button to open the activity.


abhinavraj23:
@CloudyPadmal In this pr I have just created a new multimeter activity but the layout is still the original one, it is yet to be changed, so you can add the work in progress tag, I will update this pr or else I can do that in my subsequent other PRs.

app/src/main/res/layout/activity_multimeter.xml
```diff
@@ -0,0 +1,368 @@
...
+                    android:id="@+id/button_control_read1"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:text="@string/resistance"
+                    />
+
+                <TextView
+                    android:id="@+id/tv_control_read1"
+                    android:layout_width="0dp"
+                    android:layout_height="40dp"
```
Vikum94: You can add these files to the dimensions file and call it from there. You have done it correctly at some places.

PR #938, opened 2018-05-29
chore: removed device tests from travis

### LINKED ISSUE #902: Consider reducing UI tests
There are only 2 UI tests in the app right now. It means that the test coverage in UI tests is small. Hence, the huge disadvantage of about 10 minute build time and waiting for emulator outweighs the advantage of having such small set of tests.

Either UI tests should cover lot of cases (not recommended as unit tests should handle most of this) or UI tests should be disabled and replaced with counter unit tests which will run in milliseconds rather than 20 minutes

One of the test is acually a dummy test.
Right now navigation of app is tested via UI tests. I would say that this is not such an important test but can be easily replicated in the abstracted unit test. If not, I recommend roboelectric to do the same test on JVM. Although not as fast as a unit test, it will run in 1-2 seconds as compared to current situation.

This can be taken up after crucial bug fixes and features but will significantly improve the build time and hence development iterations

CloudyPadmal:
Fixes #902

Changes: 
- Removed device connected tests to improve build time.

Screenshots for the change: 
> Not available

APK for testing: 
> Not applicable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,49 +1,16 @@
 language: android
 
-jdk: 
- - oraclejdk8
+sudo: false
 
-env:
-  global:
-    - ANDROID_API_LEVEL=27
-    - ANDROID_API_LEVEL_22=22
-    - ANDROID_BUILD_TOOLS_VERSION=27.0.3
-    - ANDROID_ABI=armeabi-v7a
-    - ANDROID_TAG=google_apis
-    - ANDROID_TARGET=android-27
-    - ADB_INSTALL_TIMEOUT=20 # minutes (2 minutes by default)
+jdk: oraclejdk8
 
 android:
-  update_sdk: true 
   components:
     - tools
     - platform-tools
-    - tools
-    - build-tools-$ANDROID_BUILD_TOOLS_VERSION
-    - android-$ANDROID_API_LEVEL
-    - android-$ANDROID_API_LEVEL_22
-    # For Google APIs
-    - addon-$ANDROID_TAG-google-$ANDROID_API_LEVEL
-    # Google Play Services
-    - extra-google-google_play_services
-    # Support library
-    - extra-android-support
-    # Latest artifacts in local repository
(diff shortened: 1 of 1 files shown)


iamareebjamal:
@CloudyPadmal Since we don't need avd dependencies, travis config can be lot smaller now. Check this out 

```
language: android
sudo: false
jdk:
    - oraclejdk8
android:
  components:
    - tools
    - android-27
    - build-tools-27.0.3
    - platform-tools
before_cache:
  - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
cache:
    directories:
      - "${TRAVIS_BUILD_DIR}/gradle/caches/"
      - "${TRAVIS_BUILD_DIR}/gradle/wrapper/dists/"
      - "$HOME/.gradle/caches/"
      - "$HOME/.gradle/wrapper/"
install:
  - echo yes | ${ANDROID_HOME}/tools/bin/sdkmanager "platforms;android-27"
script:
  - ./gradlew build
  - ./gradlew testDebugUnitTestCoverage
after_success:
    - bash <(curl -s https://codecov.io/bash)
    - bash scripts/prep-key.sh
    - bash scripts/update-apk.sh
```

.travis.yml
```diff
@@ -54,19 +21,14 @@ cache:
...
- - echo -e "\n84831b9409646a918e30573bab4c9c91346d8abd" > "$ANDROID_HOME/licenses/android-sdk-preview-license"
- - echo 'count=0' > /home/travis/.android/repositories.cfg
+install:
+    - echo yes | ${ANDROID_HOME}/tools/bin/sdkmanager "platforms;android-27"
 
 script:
   - ./gradlew build --stacktrace
 
 after_success:
+  - bash <(curl -s https://codecov.io/bash)
```
iamareebjamal: This isn't needed here

CloudyPadmal: Ahh yes! :D

PR #990, opened 2018-06-07
refactor: reduced fluctuation of resistance in multimeter

### LINKED ISSUE #976: Multimeter shows various values for a single resistance
**Actual Behaviour**

Multimeter shows various values for a single resistance  when clicked multiple time(as every resistance has some tolerance level)

**Expected Behaviour**

It should show a single value which should be average of 10 values taken.

**Steps to reproduce it**

Run an iterative loop 10 times and calculate the average.

**LogCat for the issue**

Not Applicable

Yes

abhinavraj23:
Fixes #976 

Changes:added a loop which calculates the average resistance taking 20 values, thus gives a stable value

Screenshots for the change: 
multimeter-_resistance_changed

APK for testing: [Compress the *.apk file into a rar or zip file and upload it here]
[app-debug.apk.zip](https://github.com/fossasia/pslab-android/files/2080492/app-debug.apk.zip)

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/activity/MultimeterActivity.java
+++ b/app/src/main/java/org/fossasia/pslab/activity/MultimeterActivity.java
@@ -58,21 +58,32 @@ public class MultimeterActivity extends AppCompatActivity {
             public void onClick(View view) {
                 if (scienceLab.isConnected()) {
                     DecimalFormat resistanceFormat = new DecimalFormat("#.##");
-                    Double resistance = scienceLab.getResistance();
+                    Double resistance;
+                    Double avgResistance = 0.0;
+                    int loops = 20;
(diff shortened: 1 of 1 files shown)


abhinavraj23:
@CloudyPadmal The time taken to complete the loop is not that much so the user can change the resistance. But if still the user changes the resistance it will end the loop and show infinite resistance :+1:

abhinavraj23:
@CloudyPadmal @harsh-2711 @Vikum94 @Avjeet Please review :+1:

CloudyPadmal:
It still fluctuates right? :)

abhinavraj23:
@CloudyPadmal Yes a bit of fluctuation is still there but that is very less as compared to previous. Fluctuations can be further reduced if we increase the loop number but that would result in slower process. What do you think?

PR #1036, opened 2018-06-14
feat:multimeter now saving its state

### LINKED ISSUE #1035: Multimeter not saving its state
**Actual Behaviour**

Currently the multimeter state is not being saved and is refreshed as soon as the instrument is closed

**Expected Behaviour**

The multimeter state should be saved till the time user presses the reset button or closes the app.

**Steps to reproduce it**

Add shared preferences in the multimeter activity

**LogCat for the issue**

Not applicable

**Screenshots of the issue**

Not applicable

Yes

abhinavraj23:
Fixes #1035 

**Changes**:
Added shared preferences in the multimeter activity

**APK for testing**: 
[app-debug.apk.zip](https://github.com/fossasia/pslab-android/files/2102402/app-debug.apk.zip)

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/activity/MultimeterActivity.java
+++ b/app/src/main/java/org/fossasia/pslab/activity/MultimeterActivity.java
@@ -10,9 +10,7 @@ import android.view.View;
 import android.widget.Button;
 import android.widget.CheckBox;
 import android.widget.ImageView;
-import android.widget.Spinner;
 import android.widget.TextView;
-import android.widget.Toast;
 
 import org.fossasia.pslab.R;
 import org.fossasia.pslab.communication.ScienceLab;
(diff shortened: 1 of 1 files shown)


abhinavraj23:
@CloudyPadmal @harsh-2711 @Avjeet @Vikum94 Please review

app/src/main/java/org/fossasia/pslab/activity/MultimeterActivity.java
```diff
@@ -127,6 +135,11 @@ public void onState(int state) {
         reset.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
+                saveData("", "");
```
CloudyPadmal: Any specific reason to use blank strings here?

abhinavraj23: @CloudyPadmal Through the reset button the text-box should be made blank and should remain blank thus saving the blank string is also necessary or otherwise we have to clear the shared preferences, that can also be done.

abhinavraj23:
@harsh-2711 @Vikum94 @Avjeet Please review

app/src/main/java/org/fossasia/pslab/activity/MultimeterActivity.java
harsh-2711: I think you can use a key for this purpose and then use if statement to check the state of shared preference because empty string doesn't make any sense.

Avjeet:
@abhinavraj23 You could clear the value of the preferences using **multimeter_data.edit().clear().commit()**; and then return the default value when you read them.

abhinavraj23:
@Avjeet Changes done :+1:

PR #1126, opened 2018-06-23
feat : added dynamic selection of SQ pins in PWM mode.

### LINKED ISSUE #1100: the dynamic selection of the SQ pins for PWM pins is not there
**Actual Behaviour**

Currently in the PWM mode all the SQ pins are activated for testing purposes.

**Expected Behaviour**

The user should be able to activate or deactivate the SQ pins when producing the PWM signal

**Steps to reproduce it**

WaveGenerator -> Implement the functionality of Dynamic activation of SQ pins while producing the PWM signal

Yes

Avjeet:
Fixes #1100 

**Changes**:
 - When used in PWM mode the SQ pins for which the value has been set will automatically be activated  
    for output

**APK for testing**: 
[app-debug.zip](https://github.com/fossasia/pslab-android/files/2129936/app-debug.zip)

### DIFF
--- a/app/src/main/java/org/fossasia/pslab/activity/WaveGeneratorActivity.java
+++ b/app/src/main/java/org/fossasia/pslab/activity/WaveGeneratorActivity.java
@@ -122,7 +122,6 @@ public class WaveGeneratorActivity extends AppCompatActivity {
     @BindView(R.id.btn_view)
     Button btnView;
 
-
     //bottomSheet
     @BindView(R.id.bottom_sheet)
     LinearLayout bottomSheet;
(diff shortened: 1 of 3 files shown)


Avjeet:
@CloudyPadmal @Vikum94 @harsh-2711 @abhinavraj23 plz only review the second commit , first one is from previous PR.

app/src/main/res/values/colors.xml
```diff
@@ -13,6 +13,7 @@
     <color name="back_circle_color">#EDEDED</color>
     <color name="indicator_color">#d32f2f</color>
     <color name="progress_secondary_color">#EEEEEE</color>
+    <color name="purple_btn">#6002ee</color>
```
CloudyPadmal: @Avjeet can purple cross match with red? Besides on a black background it's hard to see a purple letters and what if we use the same red as in our theme?

Avjeet:
@CloudyPadmal well red will also be not visible on black background so I used Green color.

Avjeet:
@CloudyPadmal well, the orange is when u press the button to change its properties like duty or frequency.
The blue color will only be used when you are in PWM mode and you long press the SQ pin, that will activate that pin, the blue coloring of the SQ pins on monitor is analogous to the SQ checkboxes in initial layout.

CloudyPadmal:
@Avjeet I don't know, doesn't it sound too complicated? So many settings and color changes for a simple wave generator. Can't we have a simple implementation? Do you think a school kid will be able to go through all that complicated setting up process? Long pressings; mode switching... @harsh-2711 @abhinavraj23 @Vikum94 please review this!

Vikum94:
I agree with @CloudyPadmal I also didn't get why a long press is needed. Can't we use the set button for that also @Avjeet

Avjeet:
@CloudyPadmal Ok I will remove the long press and the channels for which the value is set will be activated and for those, no value has been set will remain deactivated.

PR #1708, opened 2019-05-12
fix: App crash when back button is pressed in Import Log

### LINKED ISSUE #1707: Bug: App crashes when back is pressed with no file is selected in import log menu
**Actual Behaviour**
App crashes when back button(No file is selected) is pressed after selecting import log from logged data menu.
**Expected Behaviour**
App should return to the logged data activity instead of crashing.
**Steps to reproduce it**
Open app -> Logged data from navigation menu -> Import Log from top right menu -> select any one option -> without selecting any file from the opened intent, press back.
**Screenshots of the issue**

Yes, I would like to work.

Smit2010:
Fixes #1707 

### DIFF
--- a/app/src/main/java/io/pslab/activity/DataLoggerActivity.java
+++ b/app/src/main/java/io/pslab/activity/DataLoggerActivity.java
@@ -221,11 +221,14 @@ public class DataLoggerActivity extends AppCompatActivity {
     @Override
     protected void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {
         if (requestCode == 100) {
-            Uri uri = data.getData();
-            String path = uri.getPath();
-            path = path.replace("/root_path/", "/");
-            File file = new File(path);
-            getFileData(file);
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/pslab/activity/DataLoggerActivity.java
```diff
@@ -221,11 +221,13 @@ private void selectFile() {
     @Override
     protected void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {
         if (requestCode == 100) {
-            Uri uri = data.getData();
-            String path = uri.getPath();
-            path = path.replace("/root_path/", "/");
-            File file = new File(path);
-            getFileData(file);
+            if (data != null) {
```
adityastic: Never check for data is null, check for resultCode is ok

adityastic: And add a toast in else which states no file selected

PR #1790, opened 2019-06-11
Fix #1789 import log/ playback working fine for wave generator

### LINKED ISSUE #1789: Add feature to import and play back saved wave configurations
**Actual Behaviour**
Currently no feature to see or retrieve saved wave configs

**Expected Behaviour**
User should be able to playback saved configs like other instruments

Yes

neel1998:
Fixes #1789 

**Changes**: User can playback saved wave configurations

**APK for testing**: 
[wave_generator_import.zip](https://github.com/fossasia/pslab-android/files/3278000/wave_generator_import.zip)

@CloudyPadmal  @ho-dor  @mariobehling have a look at this

### DIFF
--- a/app/src/main/java/io/pslab/activity/WaveGeneratorActivity.java
+++ b/app/src/main/java/io/pslab/activity/WaveGeneratorActivity.java
@@ -1,10 +1,13 @@
 package io.pslab.activity;
 
 import android.annotation.SuppressLint;
+import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.graphics.drawable.Drawable;
+import android.location.Location;
+import android.location.LocationManager;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
@@ -17,6 +20,7 @@ import android.support.v7.app.AlertDialog;
 import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.TooltipCompat;
 import android.util.DisplayMetrics;
+import android.util.Log;
 import android.view.GestureDetector;
 import android.view.LayoutInflater;
 import android.view.MotionEvent;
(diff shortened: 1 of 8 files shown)


mariobehling:
I am unclear about the UI. Where are the record, play, stop, pause buttons?

neel1998:
@mariobehling  there won't be any play/pause buttons. Wave generator just set configurations for the waves like frequency and phase. Once user has set some configuration, he can view the wave in oscilloscope. Now if user wishes to use those configuration once again, he can save those configs in a csv file. Now whenever user opens a log from the data logger activity for the wave generator, the values of the respective wave would automatically be set to those stored in the log, Hence user can reuse the configs. Using the import log function, user can also import the configs shared by other users.

app/src/main/java/io/pslab/activity/WaveGeneratorActivity.java
```diff
@@ -201,6 +221,11 @@ protected void onCreate(Bundle savedInstanceState) {
 
         removeStatusBar();
 
+        realm = LocalDataLog.with().getRealm();
+        gpsLogger = new GPSLogger(this,
```
cweitat: I don't understand why do we need to plot coordinates for saving a wave generator configuration? If oscilloscope has it I could understand but wave generator?

neel1998: I just added this to have consistency and also when we display the log in data logger activity there is an option to display coordinates on the map for all instruments, that why i added it here as well

cweitat: secondly, if we don't have the map data, the map icon should change to another icon like the open folder icon.

neel1998: That can be done. But i Think we can make this changes in a separate PR, since those changes will be applied to all the instruments. I don't see any harm right now to keep coordinates in wave generator

neel1998:
@CloudyPadmal  @cweitat  can you please review and merge this? Thanks

PR #1800, opened 2019-06-16
continuous cursor blinking fixed

### LINKED ISSUE #1784: Robotic Arm: Cursor always blinking
There is a constantly blinking cursor in the number area of the degrees. Please only activate the cursor if the user activate the area. If the use clicks outside of the area the cursor should disappear and the area should be no longer highlighted.

ho-dor:
Fixes #1784 

**Changes**:Cursor not blinking anymore.

**Screenshot/s for the changes**: N/A

**APK for testing**: 

[cursor_blink.zip](https://github.com/fossasia/pslab-android/files/3297687/cursor_blink.zip)

### DIFF
--- a/app/src/main/java/io/pslab/activity/RoboticArmActivity.java
+++ b/app/src/main/java/io/pslab/activity/RoboticArmActivity.java
@@ -188,6 +188,7 @@ public class RoboticArmActivity extends AppCompatActivity {
                 } else {
                     degreeText1.setText(String.valueOf((int) (i * 3.6)));
                 }
+                degreeText1.setCursorVisible(false);
             }
 
             @Override
@@ -210,6 +211,7 @@ public class RoboticArmActivity extends AppCompatActivity {
                 } else {
                     degreeText2.setText(String.valueOf((int) (i * 3.6)));
                 }
+                degreeText2.setCursorVisible(false);
             }
 
             @Override
@@ -232,6 +234,7 @@ public class RoboticArmActivity extends AppCompatActivity {
                 } else {
                     degreeText3.setText(String.valueOf((int) (i * 3.6)));
                 }
+                degreeText3.setCursorVisible(false);
             }
 
             @Override
@@ -254,6 +257,7 @@ public class RoboticArmActivity extends AppCompatActivity {
                 } else {
                     degreeText4.setText(String.valueOf((int) (i * 3.6)));
                 }
+                degreeText4.setCursorVisible(false);
             }
 
             @Override
@@ -322,6 +326,13 @@ public class RoboticArmActivity extends AppCompatActivity {
             }
         });
 
+        degreeText1.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                degreeText1.setCursorVisible(true);
+            }
+        });
+
         degreeText2.setOnEditorActionListener(new TextView.OnEditorActionListener() {
             @Override
             public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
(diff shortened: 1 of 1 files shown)


ho-dor:
@neel1998 check the apk

ho-dor:
> @ho-dor Can we show the cursor when the user selects an edit box. Also when user adds value out of the range of 0 - 360 the value isn't automatically getting set to zero. Please check

Actually , it is a work around where I have removed the cursor , to indicate that the edittext is active , the backgroundtint is red when in focus but if I show the cursor during this time , it will blink continuously until the done is pressed in soft keyboard which is the issue.

app/src/main/java/io/pslab/activity/RoboticArmActivity.java
```diff
@@ -318,10 +318,18 @@ public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
...
                 }
+                degreeText1.clearFocus();
                 return false;
             }
         });
 
+        degreeText1.setOnFocusChangeListener(new View.OnFocusChangeListener() {
+            @Override
+            public void onFocusChange(View v, boolean hasFocus) {
+                degreeText1.setCursorVisible(false);
```
CloudyPadmal: What if we make cursor visibility false in every other text? As an example; here it'll be:
degreeText1.setCursorVisible(true);
degreeText2.setCursorVisible(false);
degreeText3.setCursorVisible(false);
degreeText4.setCursorVisible(false);

Will it work? Just try it and see :thinking:

ho-dor: But the cursor is already false for all other texts if I have clicked on one of them. The issue is when we don't press the Done button on the soft keyboard and so the focus stays on that one text and the cursor blinks continuously until the Done is pressed.

CloudyPadmal: `setOnFocusChangeListener` will be fired for only two views. One that is getting the focus and the one that is losing focus. Two textviews will not get this method called.

ho-dor:
@CloudyPadmal @neel1998 please check , I have updated the apk.

app/src/main/java/io/pslab/activity/RoboticArmActivity.java
```diff
@@ -248,11 +257,14 @@ public void onStopTrackingTouch(SeekArc seekArc) {
...
             @Override
             public void onProgressChanged(SeekArc seekArc, int i, boolean b) {
+                degreeText4.requestFocus();
                 if (editEnter) {
                     degreeText4.setText(String.valueOf(degree));
                     editEnter = false;
+                    degreeText4.setCursorVisible(false);
                 } else {
                     degreeText4.setText(String.valueOf((int) (i * 3.6)));
+                    degreeText4.setCursorVisible(false);
```
CloudyPadmal: Since this is in both if and else we can take it out and have only one statement

app/src/main/java/io/pslab/activity/RoboticArmActivity.java
```diff
@@ -226,11 +232,14 @@ public void onStopTrackingTouch(SeekArc seekArc) {
         seekArc3.setOnSeekArcChangeListener(new SeekArc.OnSeekArcChangeListener() {
             @Override
             public void onProgressChanged(SeekArc seekArc, int i, boolean b) {
+                degreeText3.requestFocus();
```
CloudyPadmal: If we're making the cursor invisible, let's remove this request focus part and keep setcursorvisible false outside if else statement.

ho-dor:
@CloudyPadmal please check

neel1998:
@CloudyPadmal  i think this cab be merged

PR #1836, opened 2019-06-28
[WIP]Google maps option added

### LINKED ISSUE #1739: Option to use OSM or Google Map
**Actual Behaviour**
Currently we are only using OSM to plot locations. However Google maps is not open source so FDroid does not allow to release 

**Expected Behaviour**
Add option for OSM or Google Map on Play Store app and just OSM in FDroid.

**Steps to reproduce it**
N/A

Yes

ho-dor:
Fixes #1739

**Changes**: Created two flavours fdroid and playstore and changes made in playstore flavour

- [ ] No modifications done at the end of resource files `strings.xml`, `dimens.xml` or `colors.xml`

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -31,6 +31,15 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+    flavorDimensions 'default'
+    productFlavors {
+        fdroid {
+            dimension = 'default'
+        }
+        playstore {
+            dimension = 'default'
+        }
+    }
 }
 
 dependencies {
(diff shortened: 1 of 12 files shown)


ho-dor:
This manifest merger error is really frustrating. No idea what is going wrong.

neel1998:
Do we really need to create a whole new sensorListAdapter class?

ho-dor:
Yes. That is where the main decision is taken between OSM and Google Map and if I don't create a new SensorListAdapter and add GoogleMap condition in main itself then there will be error in fdroid build.
If you have any better ideas, please share.

neel1998:
Can't we have a separate class with methods to plot map data...And call that class from SensorListAdapter... So we only need to create 2 version of that map class , since rest of logic of SensorListAdaptet class is same...That would reduce the redundancy a lot.Does it make sense?

neel1998:
@CloudyPadmal please share your views

ho-dor:
@neel1998 something like now?

neel1998:
@ho-dor yes. Doesn't make more sense now?. Will have to look for Travis still

ho-dor:
It's that same issue , manifest merge.

PR #1905, opened 2019-08-01
create config file working fine

### LINKED ISSUE #1882: PSLab SD Card configuration
**Actual Behaviour**

In order to use PSLab without using a device (say, mount on a weather station), a configuration file is needed to save in SD card so that the PSLab can read the setting from this file and do the processing and logging

**Expected Behaviour**

Have these properties in the config file in YAML format
- File Name (Optional)
- Instrument Name
- Logging Interval
- Parameters needed to be log (Depending on the instrument, Oscilloscope:: CH1, CH2 etc..)

**Steps to reproduce it**

> Not available

**LogCat for the issue**

> Not available

**Screenshots of the issue**

> Not available

Issue for GSoC

neel1998:
Fixes #1882 
**Changes**: User can create config files

**APK for testing**: 
[cc.zip](https://github.com/fossasia/pslab-android/files/3456043/cc.zip)

@CloudyPadmal  my local branch for previous PR got messy because of an error while using BFG. Regarding storing files directly on SD card, i did looked up to many docs and stackoverflow, but apparently Android doesn't allow apps other then system apps to store files directly on /sdcard/ folder. We can store files on the app specific folder on sd card which is /sdcard/Android/data/<package>/files, bu i asume in our case that wont make much sense. A better option would be storing the file on local storage and then user can copy the file easily to sdcard folder. Still i have asked a question regarding this on stackoverflow and if i get some solution to do this, i will surely incorporate that. Thanks

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -27,6 +27,7 @@
         android:theme="@style/AppTheme"
         tools:replace="android:icon">
         <uses-library android:name="org.apache.http.legacy" android:required="false"/>
+        <activity android:name=".activity.CreateConfigActivity"/>
         <activity
             android:name=".activity.ThermometerActivity"
             android:screenOrientation="portrait"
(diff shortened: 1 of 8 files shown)


CloudyPadmal:
Generate the config file in PSLab folder; where it'll be easier to find without browsing into multiple directories inside.

neel1998:
I am generating the file in the PSLab/ folder only not in instrument sub directory

PR #1927, opened 2019-08-13
feat: add Httphandler class

### LINKED ISSUE #1926: Add HTTPHandler to communicate with ESP
**Actual Behaviour**
Currently there is no httphandler

**Expected Behaviour**
Add a httphandler like communicationHandler to control all the http communications with esp

**Steps to reproduce it**
N/A

**LogCat for the issue**
N/A

**Screenshots of the issue**
N/A

Yes

neel1998:
Fixes #1926 

**Changes**: added http handler class 

N/A

**APK for testing**: 
N/A

### DIFF
--- a/app/src/main/java/io/pslab/communication/HttpAsyncTask.java
+++ b/app/src/main/java/io/pslab/communication/HttpAsyncTask.java
@@ -1,0 +1,43 @@
+package io.pslab.communication;
+
+import android.os.AsyncTask;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.IOException;
+
+import io.pslab.interfaces.HttpCallback;
+
+public class HttpAsyncTask extends AsyncTask<byte[], Void, Void> {
+
+    private HttpHandler mHttpHandler;
+    private HttpCallback<JSONObject> mHttpCallback;
+
+    public HttpAsyncTask(String baseIP, HttpCallback<JSONObject> httpCallback) {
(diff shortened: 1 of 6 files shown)


neel1998:
**Current Architecture:**
currentArchitecture

**New Architecture:**
newArchitecture

@CloudyPadmal as shown in the above diagram we can add a new WIFI pathway to existing architecture. I have made two new functions ```commonRead()``` and ```commoWrite()``` in ```PacketHandler``` class... This will be the only point where the two architectures will diverge. I think this architecture is quite neat, and since we are not making any changes in the ScienceLab class we won't require to change a hell lot of code for wifi communication, which would also reduce the redundancy. This is still WIP. But kindly share your views. Thanks

CloudyPadmal:
Yes this is the architecture proposed. Before going any further, better check if it works as intended with everything integrated (both ESP and USB).

neel1998:
@CloudyPadmal yes. I have already tested the new architecture for USB path and it is working as expected. So that's a good sign.

CloudyPadmal:
Let's use labels instead of change of titles

neel1998:
@CloudyPadmal  I have tested the wifi path on a dummy server. The following path in the architecture to communicate over wifi was able to make get/post request to a server,

ESPFragment -> Initial Connection -> HttpAsyncTask -> HttpHandler -> HttpCallback

I think we can merge this basic architecture and then develop further on top of it.

PR #1987, opened 2019-10-15
development

rsrohan:
Redundant and Unused Variables excluded. Lambda implemented on listeners.


CloudyPadmal:
These kind of PRs are highly discouraged. It's just removing code pointed out by android studio code inspection. Build might pass, but there will be tons of bugs when used with a real device.

PR #2025, opened 2019-12-12
Added missing screenshots

Hasala123:
I added 3 new screenshots that were missing in the ReadMe file which include the Sensors view, Gas Sensor view and the Dust Sensor view. I have attached a screenshot of the result in my repository[
3 new screenshots
]


mariobehling:
It would be better if the instruments would actually show data being displayed. We don't see much going on.

Hasala123:
I dont actually have any of the sensors, but i can edit the screenshots to look like they have a reading.

Hasala123:
I added the charts using an existing screenshots. I left the values as zero because the existing instrument values were also zero even with a generated chart.
3 new screenshots

techyay:
@CloudyPadmal How can @Hasala123 get a reading with data? Let me merge this and we try to get to the next step in a follow up PR.

PR #2084, opened 2020-01-19
Fixed About page crash issue

### LINKED ISSUE #2083: bug: About page keeps Shifting theme which leads to crashes in android 9 and 10
**Actual Behaviour**

Currently the application's About page randomly keeps shifting themes in android 9 and 10 *(maybe also 8 but I couldn't verify this)* when system theme is set to dark and application is switched and also causes crashes in the case of my android 10 testings. 

Screenshot_20200117-185105
Screenshot_20200117-185120
Screenshot_20200117-185126

**Expected Behaviour**

The application theme shouldn't switch in fly it should remain the same and should never cause crashes

**Steps to reproduce it**
- Install pslab into a Android 9 or 10 phone with dark theme on
- go to about page ( you should see the dark theme of the about page)
- now switch application and come back to pslab

**Resources of the issue**
in android 9
ezgif com-optimize

in Android 10
ezgif com-video-to-gif

**(it doesn't crash everytime but it does in 1/4 scenario in my avd testings)**

Yes I would like to work on the issue

Sid911:
Fixes #2083 

**Changes**: 
- **Removed the simulate Day Night Function** : The About section is powered by  https://github.com/medyo/android-about-page this package which unfortunately uses` App Compat Day Night` as a parent style which directly takes the information from system variable to change the theme **actively**. The function `simulateDayNight` counters that by forcing the app theme to change back to light. This creates a loop you can see in bottom gif, in android 10 of my testing in avd even shows more frequent crashes.

ezgif com-optimize

Before : 
ezgif com-video-to-gif

Now: 
ezgif com-video-to-gif (1)

**APK for testing**: [Compress the `app-debug.apk` file into a `<feature>.rar` or `<feature>.zip` file and upload it here]

[aboutPageCrash.zip](https://github.com/fossasia/pslab-android/files/4083747/aboutPageCrash.zip)

### DIFF
--- a/app/src/main/java/io/pslab/fragment/AboutUsFragment.java
+++ b/app/src/main/java/io/pslab/fragment/AboutUsFragment.java
@@ -39,7 +39,6 @@ public class AboutUsFragment extends Fragment {
     public View onCreateView(LayoutInflater inflater, ViewGroup container,
                              Bundle savedInstanceState) {
         View view = inflater.inflate(R.layout.fragment_about_us, container, false);
-        simulateDayNight(0);
         ButterKnife.bind(this, view);
         View aboutPage = new AboutPage(getActivity())
                 .isRTL(false)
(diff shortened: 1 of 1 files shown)


Sid911:
I tired fixing the darktheme but the android-about-page repo's last production build is more than a year old and there is almost no way this can be fixed other than fiddling with the jar files 😅, *( I did find something here https://github.com/medyo/android-about-page/issues/113)* ... also the issue was about crash and switching of theme on the fly so I left the dark theme as it is 😇

Sid911:
I will try to fix that in next pr if possible

Sid911:
I already worked on dark mode system which is much more flexible than all these things and is kind of future proof considering its considered under https://developer.android.com/jetpack/androidx .. and doesn't rely on android support libraries (which will soon be unsupported) https://github.com/fossasia/pslab-android/issues/2065 but that doesn't seem to be our priority now 😅

Sid911:
@cweitat please review this 😄

kumuditha-udayanga:
LGTM!

Sid911:
can this be merged?

app/src/main/java/io/pslab/fragment/AboutUsFragment.java
```diff
@@ -66,7 +65,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
         return super.onOptionsItemSelected(item);
     }
 
-    private void simulateDayNight(int currentSetting) {
+/*    private void simulateDayNight(int currentSetting) {
```
CloudyPadmal: Rather than commenting out, let's remove this section. If we want to refer back, we can always use git history

Sid911:
@CloudyPadmal Done the changes 😇

adityastic:
@CloudyPadmal RTM

app/src/main/java/io/pslab/fragment/AboutUsFragment.java
Sid911: Done @CloudyPadmal

mariobehling:
@Sid911 Please resolve conflicts.

Sid911:
@CloudyPadmal @mariobehling Done

PR #2089, opened 2020-01-21
Create gci.md

boy-who-codes:
This is only for GCI 2020


mariobehling:
Sorry, but this is considered spam usually in projects. Please dont reply and dont do this again.

PR #2112, opened 2020-05-17
Fix: Logged data not shown for Gas Sensor.

### LINKED ISSUE #1984: Logged data not shown for gas sensor
**Actual Behaviour**

No logged data shown after recording in gas sensor.

**Expected Behaviour**

Recorded data should be shown after recording.

**Steps to reproduce it**

Record data on gas sensor and open logged data.

**LogCat for the issue**

N/A

**Screenshots of the issue**

Yes.

VIKAS9899:
Fixes #1984

**Changes**: 
Entry would be created for case when device is not created to Gas sensor.

**Video for the change**: https://github.com/VIKAS9899/LogRepo/raw/master/Upload/2020_05_17_14_42_40.mp4

**APK for testing**: https://github.com/VIKAS9899/LogRepo/raw/master/Upload/pslab_1984.apk

### DIFF
--- a/app/src/main/java/io/pslab/fragment/GasSensorDataFragment.java
+++ b/app/src/main/java/io/pslab/fragment/GasSensorDataFragment.java
@@ -368,9 +368,10 @@ public class GasSensorDataFragment extends Fragment {
     }
 
     private void visualizeData() {
+        double ppmValue = 0d;
         if (scienceLab.isConnected()) {
             double volt = scienceLab.getVoltage("CH1", 1);
-            double ppmValue = (volt / 3.3) * 1024.0;
+            ppmValue = (volt / 3.3) * 1024.0;
             gasValue.setText(String.format(Locale.getDefault(), "%.2f", ppmValue));
             gasSensorMeter.setWithTremble(false);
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/pslab/fragment/GasSensorDataFragment.java
```diff
@@ -397,6 +397,7 @@ private void visualizeData() {
                 mChart.invalidate();
             }
         }
+        writeLogToFile(System.currentTimeMillis(), (float) ppmValue);
```
CloudyPadmal, line 400: Do we need to cast the double to float? What if the variable is set to float in the first place?

CloudyPadmal:
Also squash commits into one and stick to the PR template.

app/src/main/java/io/pslab/fragment/GasSensorDataFragment.java
fcartegnie: https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.3

CloudyPadmal (approved):

Issue #2126, opened 2020-09-02
Faq section expandable list view not working 


CodeWithSouma:
### Expected Behaviour
It should show answer of that question 

### Screenshots of the issue

<pre><img src = "https://user-images.githubusercontent.com/38830139/91998200-8efbe800-ed58-11ea-835b-12fcc50adb33.png" width="200"/> <img src = "https://user-images.githubusercontent.com/38830139/91999851-7c82ae00-ed5a-11ea-8766-30423551280b.png" width="200" /> <img src="https://user-images.githubusercontent.com/38830139/91999994-a5a33e80-ed5a-11ea-8f1f-7952f27d0996.png" width="200"/> <img src="https://user-images.githubusercontent.com/38830139/92000650-5e697d80-ed5b-11ea-9357-7c458300f778.png" width="200"/></pre>

### Gif 
<pre>
                                        <img src="https://user-images.githubusercontent.com/38830139/92000818-953f9380-ed5b-11ea-9cf6-de7d5361b601.gif" width="200" /> </pre>

ritmojs:
I want to work on this issue

CodeWithSouma:
> I want to work on this issue

I already **fixed** it. and send **PR** #2127

nileshteji:
@CodeWithSouma  the pr you mentioned is not mergest . Are you still working on this issue?Thank you

CodeWithSouma:
> Are you still working on this issue?

Yes i am woking on this issue

PR #2127, opened 2020-09-02
Faq error 

CodeWithSouma:
# Fix Faq expandable listview error #2126
* use expandable listView
* create a custom adapter 
* set that adapter in expandable listView


app/src/main/java/io/pslab/adapters/CustomAdapter.java
```diff
@@ -0,0 +1,96 @@
...
+import android.webkit.WebView;
+import android.widget.BaseExpandableListAdapter;
+import android.widget.TextView;
+
+import java.util.HashMap;
+import java.util.List;
+
+import io.pslab.R;
+
+public class CustomAdapter extends BaseExpandableListAdapter {
```
CloudyPadmal, line 18: Use meaningful names for variables (Instead of "CustomAdapter" use something that makes sense). This applies to the rest of PR.

CloudyPadmal (changes requested):

app/src/main/java/io/pslab/adapters/CustomAdapter.java
CodeWithSouma: ## Changed CustomAdapter name
Refactored `CustomAdapter` to `FaqListAdapter`

app/src/main/java/io/pslab/fragment/FAQFragment.java
```diff
@@ -25,128 +31,34 @@ public static FAQFragment newInstance() {
...
-        super.onViewCreated(view, savedInstanceState);
-        ExpandableListView listView;
-
-        listView = (ExpandableListView) view.findViewById(R.id.expListView);
-        listView.setAdapter(new ExpandableListAdapter(questions, answers));
-        listView.setGroupIndicator(null);
-
-    }
-
-    public class ExpandableListAdapter extends BaseExpandableListAdapter {
```
CloudyPadmal, line 55: Since we are not using such adapter class anywhere else in the app, should we really take it out from here and create as a separate class?

CloudyPadmal (changes requested):

app/src/main/java/io/pslab/fragment/FAQFragment.java
CodeWithSouma: > Since we are not using such adapter class anywhere else in the app, should we really take it out from here and create as a separate class?

Sir if you want i can put this adapter in FAQFragment.java as a nested class rather than crreate a seperate adapter but i think  this type of implementation hard to maintain

CloudyPadmal: I wonder why it's hard to maintain if the only place the class is used is inside FAQFragment. What else changed in this PR except moving the inner class to a separate outer class?

CodeWithSouma: actually previous adapter has some problem when we click on expandable listview the app was crash. #2126  please sir check it here i provide ss and gif.Sir i also refactored code and put this adapter inside FAQFragment ok.

CloudyPadmal (commented):
I wonder what has changed in this PR to address the parent issue. What I can see is, the String arrays were replaced by Lists and HashMaps and rearranged the code here and there. @CodeWithSouma can you elaborate more on what was the bug and what was the solution? Thanks..

CodeWithSouma:
> I wonder what has changed in this PR to address the parent issue. What I can see is, the String arrays were replaced by Lists and HashMaps and rearranged the code here and there. @CodeWithSouma can you elaborate more on what was the bug and what was the solution? Thanks..

line no 120  ` Html.FROM_HTML_MODE_COMPACT` it should not use it causes the app crash. I preferred hashMap in expandable listView so i also refactor the code.If you dont like this i can it and fix the bug.

CloudyPadmal:
> > I wonder what has changed in this PR to address the parent issue. What I can see is, the String arrays were replaced by Lists and HashMaps and rearranged the code here and there. @CodeWithSouma can you elaborate more on what was the bug and what was the solution? Thanks..
> 
> line no 120 ` Html.FROM_HTML_MODE_COMPACT` it should not use it causes the app crash. I preferred hashMap in expandable listView so i also refactor the code.If you dont like this i can change it and fix the bug.

Yes. Let's modify only the parts related to the crash. If you want to change the code base, we can open an issue and work it on a separate PR.

CodeWithSouma:
### Sir i close this PR and please check out this PR #2138 . I fix this issue as you want please check out and merge this PR #2138

PR #2177, opened 2021-01-21
feat: rrror message displayed in IP address text box

### LINKED ISSUE #2176: Keyboard hiding the snackbar output in USB Connection.
**Actual Behaviour**

When the user enters the IP Address of ESP module the error message is hidden behind the keyboard .User is able to see the snackbar output only if we press back button to remove keyboard 

**Expected Behaviour**

The error message should be visible to the user.

**Steps to reproduce it**

The edittext that is used to enter the IP address is replaced by materialdesign TextInputLayout ,so that we can view the error messages near the edittext instead of snackbar from bottom.

**Screenshots of the issue**

https://user-images.githubusercontent.com/70392921/105354591-c50e0b80-5c16-11eb-8374-7caa141a030c.mp4

asaikarthikeya:
Fixes #2176 

**Changes**: 

"Please enter valid IP address "  message showed to user in 2 ways depending on the entered state (if nothing entered and entered wrong)

https://user-images.githubusercontent.com/70392921/105355477-39957a00-5c18-11eb-86b0-98bc816566aa.mp4

### DIFF
--- a/app/src/main/res/layout/fragment_esp.xml
+++ b/app/src/main/res/layout/fragment_esp.xml
@@ -2,21 +2,32 @@
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:padding="@dimen/esp_fragment_padding">
 
-    <EditText
-        android:id="@+id/esp_ip_edit_text"
-        android:layout_width="wrap_content"
+    <android.support.design.widget.TextInputLayout
+        android:id="@+id/input_layout_password"
+        android:layout_width="match_parent"
         android:layout_height="wrap_content"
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/fragment_esp.xml
```diff
@@ -2,21 +2,32 @@
...
     android:layout_width="match_parent"
     android:layout_height="match_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:padding="@dimen/esp_fragment_padding">
 
-    <EditText
-        android:id="@+id/esp_ip_edit_text"
-        android:layout_width="wrap_content"
+    <android.support.design.widget.TextInputLayout
+        android:id="@+id/input_layout_password"
```
CloudyPadmal, line 9: Why is it a password?

CloudyPadmal (changes requested):

PR #2178, opened 2021-01-21
updated setupGuideLayout in abstract class PSLabSensor to hide image if not available in a sensor's bottomsheet guide

Ekagra:
Fixes: [empty imageview is sensors bottomsheet guide]

**Changes**: [updated setupGuideLayout in abstract class PSLabSensor to hide image if not available in a sensor's bottomsheet guide]

`Before-------------------------------------------------------------------------------------------------------------After`

<p float="left">
  
   
</p>


asaikarthikeya (commented):
In every other instrument this space is filled with a PSlab hardware related image , I think adding a image would be better since it improve regularity in app . This would be a optimal solution till we get a image. Thank you.

CloudyPadmal (changes requested):
Add an image instead of hiding the view.

CloudyPadmal:
Also try to shorten the PR title

Ekagra:
@CloudyPadmal what image should I add?

Ekagra:
The view is only hidden if there's no image to display like right now, do you want me to add a placeholder?

CloudyPadmal (approved):

PR #2207, opened 2021-03-20
Replaced getDrawable() with ResourcesCompat.getDrawable()

### LINKED ISSUE #2206: getDrawable() is deprecated and needs to be replaced
**Actual Behaviour**

getDrawable( int ) is deprecated and is used in many places in the project

**Expected Behaviour**

getDrawable() needs to be replaced, we can use ResourcesCompat.getDrawable() 

Yes

Abhinavtdk:
Fixes #2206 

**Changes**: 
There were many places where getDrawable(int) was used but it was deprecated. I replaced all these cases with ResourcesCompat.getDrawable().  

**APK for testing**: 
[replaced_getDrawable().zip](https://github.com/fossasia/pslab-android/files/6175706/replaced_getDrawable.zip)

### DIFF
--- a/app/src/fdroid/java/io/pslab/activity/MapsActivity.java
+++ b/app/src/fdroid/java/io/pslab/activity/MapsActivity.java
@@ -1,6 +1,7 @@
 package io.pslab.activity;
 
 import android.os.Bundle;
+import android.support.v4.content.res.ResourcesCompat;
 import android.support.v7.app.AppCompatActivity;
 
 import org.json.JSONArray;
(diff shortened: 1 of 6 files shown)


Abhinavtdk:
@CloudyPadmal can you review it

CodeWithSouma (changes requested):
Is it necessary to change getResources() with ResourcesCompat() ???

Abhinavtdk:
> Is it necessary to change getResources() with ResourcesCompat() ???

For now, it offers no difference. But, if we want to, we can associate the drawable resource with a particular theme. I just think it's better to use an alternative as the function is deprecated

CloudyPadmal:
If it's no difference as for now, we can leave it as it is. There are many other things need fixing. May be lets work on those.

Issue #2093, opened 2020-01-26
bug: Recording start even when device not connected


dheerajkotwani:
**Actual Behaviour**
Recording can be started even if the device is not connected to mobile or mobile does not have inbuilt feature support

**Expected Behaviour**
If any device is not connected then for that device recording should not start

**Steps to reproduce it**
Open any device which is not connected or whose sensor are not present in mobile(Gas Sensor)
Click on Record button
See the error

**Screenshots of the issue**
![recording_issue](https://user-images.githubusercontent.com/44086235/73130918-dafae780-4027-11ea-8d69-cc2ca60013fc.gif)

**Would you like to work on the issue?**
Yes

mariobehling:
Expected behavior would be a pop up "Please connect a device or choose an phone sensor (if available)" and two buttons "Configure Active Sensor" and "Cancel".

mariobehling:
In sensors that are usually present on the phone the recording should work nevertheless.

SpandanBasu:
I am a beginner. I want to work on this issue please assign me

SpandanBasu:
When the sensor is not present and the user clicks on the recording button, A pop-up dialog comes..right..but on pressing on Configure active sensor where should the app take the user?

sambhavsaxena:
Please assign this issue to me

SpandanBasu:
This issue is already fixed i opened a pull request. Try a different issue. Thanks

shubham-200315:
Hey @mariobehling , Issue is fixed and PR is merged already so i think we can close this.
Thank you.

marcnause:
Fixed in https://github.com/fossasia/pslab-android/pull/2295. A notification (SnackBar) is shown that informs the user that no device is connected.

PR #2295, opened 2022-01-14
fix : Recording starts even when device not connected.

apoorva-raj:
**Fixes** #2093 

**Changes**: 
- Created a method in the PSLabSensor.java abstract class to check whether the android device has the required sensor or not.
- If the phone doesn't have the required sensor in-built and PS Lab is not connected, data recording won't start.

**Checklist**: 
- [x] I have used resources from `strings.xml`, `dimens.xml` and `colors.xml` without hard-coding any value.
- [x] No modifications done at the end of resource files `strings.xml`, `dimens.xml` or `colors.xml`.
- [x] I have reformatted code and fixed indentation in every file included in this pull request
- [x] My code does not contain any extra lines or extra spaces.
- [x] I have requested reviews from maintainers.

### DIFF
--- a/app/src/main/java/io/pslab/models/PSLabSensor.java
+++ b/app/src/main/java/io/pslab/models/PSLabSensor.java
@@ -43,6 +43,7 @@ import io.pslab.R;
 import io.pslab.activity.DataLoggerActivity;
 import io.pslab.activity.MapsActivity;
 import io.pslab.activity.SettingsActivity;
+import io.pslab.communication.ScienceLab;
 import io.pslab.fragment.SoundMeterDataFragment;
 import io.pslab.interfaces.OperationCallback;
 import io.pslab.others.CSVLogger;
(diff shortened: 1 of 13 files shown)


CloudyPadmal (changes requested):
1. Always keep your branch updated
2. Reformat code
3. Fix issues reported by Codacy

apoorva-raj:
Fixed the problems.

app/src/main/java/io/pslab/activity/MainActivity.java
```diff
@@ -52,55 +53,76 @@
...
-
 public class MainActivity extends AppCompatActivity {
 
     private static final String TAG = "MainActivity";
-
+    private static final String TAG_DEVICE = "device";
+    private static final String TAG_INSTRUMENTS = "instruments";
+    private static final String TAG_ABOUTUS = "aboutUs";
+    private static final String TAG_PINLAYOUT = "pinLayout";
+    private static final String TAG_FAQ = "faq";
```
CloudyPadmal, line 63: There are lot of changes where they are just reshuffling content like this block. Please revert all of them back.

CloudyPadmal (changes requested):
Don't edit (reshuffle) rest of the unrelated code.

CloudyPadmal:
Also there are several commits related to gradle fixes. Update the PR and squash commits into one. https://github.com/fossasia/pslab-android#how-to-git-squash

apoorva-raj:
@CloudyPadmal  I tried to change the settings in my IDE and then reformat the code but I don't know why it is still shuffling some parts. The app is working fine though and the fixes I made are reflected there.

app/src/main/java/io/pslab/models/PSLabSensor.java
```diff
@@ -316,9 +318,12 @@ public boolean onOptionsItemSelected(MenuItem item) {
               will fire up. If user declines to give permission, don't do anything.
              */
             case R.id.record_data:
-                if (!isRecording) {
+                if (!isRecording && (!sensorNotFound() || scienceLab.isConnected())) {
```
CloudyPadmal, line 321: According to this logic, if PSLab is connected, it doesn't check the sensor test right?

CloudyPadmal (changes requested):

app/src/main/java/io/pslab/models/PSLabSensor.java
apoorva-raj: It checks but user would be able to start the recording nevertheless since PSLab is connected.

CloudyPadmal (approved):

PR #2364, opened 2023-02-06
Fix calculation of triangle wave

### LINKED ISSUE #2100: Wave Generator: triangle wave is not triangular, but cut off
Wave Generator: This issue was reported by several users. The Triangle wave is not as it is supposed to be. The triangle wave is not triangular, but cut off.

marcnause:
Fixes #2100

## Changes 
- fixed formula for calculation of triangle wave
- reformatted code
- optimized imports

## Screenshots / Recordings  
I did not change the UI, but this is what the wave looks like in the oscilloscope now.

Screenshot_20230206-225812

**Checklist**: 

### DIFF
--- a/app/src/main/java/io/pslab/communication/ScienceLab.java
+++ b/app/src/main/java/io/pslab/communication/ScienceLab.java
@@ -1,23 +1,14 @@
 package io.pslab.communication;
 
+import static org.apache.commons.lang3.math.NumberUtils.max;
+import static io.pslab.others.MathUtils.linSpace;
+
 import android.os.Handler;
 import android.os.Looper;
 import android.util.Log;
 
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.math3.analysis.polynomials.PolynomialFunction;
-import io.pslab.activity.MainActivity;
-import io.pslab.communication.analogChannel.AnalogAquisitionChannel;
-import io.pslab.communication.analogChannel.AnalogConstants;
-import io.pslab.communication.analogChannel.AnalogInputSource;
-import io.pslab.communication.digitalChannel.DigitalChannel;
-import io.pslab.communication.peripherals.I2C;
-import io.pslab.communication.peripherals.MCP4728;
-import io.pslab.communication.peripherals.NRF24L01;
-import io.pslab.communication.peripherals.RadioLink;
-import io.pslab.communication.peripherals.SPI;
-import io.pslab.fragment.HomeFragment;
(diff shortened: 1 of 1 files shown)


marcnause:
I compared my changes with the implementation in https://github.com/fossasia/pslab-python/blob/development/pslab/instrument/waveform_generator.py#L235 and it differs now. Should I revise my changes to stay closer to the Python implementation (which seems to work correctly according to a short test I conducted) or is it OK to differ as long as the functionlaity is correct?

bessman:
Hi @marcnause, and thanks for the pull request! Sorry it's taken a while to get a response.

The divergence from the pslab-python implementation is in fact a problem, I think, unless we have a clear understanding of why it is necessary. Ultimately, pslab-android and pslab-python both control the PSLab by sending it bytes over serial; if the bytes are the same, the PSLab's response will be the same.

If the current implementation of the triangle waveform in pslab-android results in different behavior than the corresponding pslab-python implementation, that must be because pslab-android and pslab-python do not send the same bytes. That is the root cause, and should be fixed.

In any case, as far as I can tell your proposed change should not affect the shape of the waveform. I don't have the android app installed at the moment so I can't test it, but I graphed what actually gets sent to the PSLab for both the current implementation and the proposed change:

Figure_1

As you can see, your change just phase-shifts the wave by 90 degrees. That should not affect the shape of the wave.

Could you post a screenshot of what the wave looks like for you without this change?

bessman:
I installed the pslab-android app and I can reproduce the bug. Still not comfortable merging this without understanding the root cause.

bessman:
The problem is that in java `-1 % 4 == -1`, while in python `-1 % 4 == 3`. Therefore, what looks like the same function actually produces different output for the same input.

I would like a comment explaining this divergence from pslab-python. Otherwise I am happy with this PR.

app/src/main/java/io/pslab/communication/ScienceLab.java
```diff
@@ -2989,7 +3038,7 @@ public void loadEquation(String channel, String function) {
                     y.add(Math.sin(x.get(i)));
                     break;
                 case "tria":
-                    y.add(abs(x.get(i) % 4 - 2) - 1);
+                    y.add(Math.abs(x.get(i) % 4 - 2));
```
bessman, line 3041: Please add a comment explaining that the modulo operator in java returns the negative remainder for negative operands, unlike python's modulo operator which always returns the positive remainder.

bessman:
On second thought, it's better to just update pslab-python to match this implementation.

PR #2413, opened 2024-05-16
fix: removed pol values and fixed gradle build files to support correct languages

### LINKED ISSUE #2410: Fix: App uses pol instead of pl for Polish Language
Please fix the issues that the app uses "pol" instead of "pl" for the Polish language:

Your App Bundle targets the following unrecognized languages: pol. The list of supported language codes can be found in the [IANA registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry). Invalid languages caused by third-party libraries can be excluded using the [resConfigs](https://developer.android.com/studio/build/shrink-code#unused-alt-resources) Gradle property.

AsCress:
Fixes #2410 
Removes the values/pol/string.xml file and updates the build.gradle.kts file to only support valid languages, as the correct file for Polish language (values/pl/strings.xml) is already present.
## Changes 
- app/src/main/res/values-pol/string.xml - File removed
- app/build.gradle.kts - resConfig property added

## Screenshots / Recordings  
N/A

**Checklist**: 


mariobehling (changes requested):
Thank you. Instead of removing the Polish language entirely, could you just rename it to "pl", please?

AsCress:
> Thank you. Instead of removing the Polish language entirely, could you just rename it to "pl", please?

I've not removed the Polish language. There is already a proper file existing for the polish language, the 'pl' file. I've just removed the redundant 'pol' one, which was causing the issue. I've specified the Polish language in the resConfig property, and the existing 'pl' file is perfectly usable.

adityastic (approved):

Issue #2173, opened 2021-01-20
Oscilloscope: After selecting PSLab MIC, In-Built-MIC does not work when PSLab device is not connected


arpoch:
**Actual Behaviour**
On selecting PSLab MIC option before the In-Built MIC option, no popup stating if a PSLab device is not connected, when selecting In-Build MIC no activity can be seen on the monitor.

**Expected Behaviour**
The PSLab MIC option should be unable when then PSLab device is not connected.

**Screenshots of the issue**
![Screenshot_20210120-125534_PSLab](https://user-images.githubusercontent.com/60058139/105141877-ca3c5f00-5b1f-11eb-8156-b32946d011be.jpg)

**Would you like to work on the issue?**
Yes

arpoch:
If have applied the changes, and the code is working fine. So should I know make a PR, can any please guide me through the procedure as it's my first PR.

![Screenshot_20210122-225917_PSLab](https://user-images.githubusercontent.com/60058139/105525065-90797d00-5d06-11eb-8d8f-a1ea8867c9f4.jpg)

asaikarthikeya:
Good , create a branch and push the changes to that branch. latter make  pull request to development branch .

asaikarthikeya:
Good. Create a new branch and push the changes to that branch. Later make pull request to development branch .

AsCress:
Superseeded by #2421 and resolved by #2422.

PR #2422, opened 2024-05-19
fix: fixed the In-Built MIC to work properly after toggling the PSLab MIC

### LINKED ISSUE #2421: After toggling the PSLab MIC, the In-Built-MIC feature doesn't work, when the PSLab device isn't connected
This issue is a proper rephrase of #2173, since the problem isn't clearly stated there.
**Actual Behaviour**
If one first toggles the PSLab MIC, the In-Built-MIC feature doesn't work after that, regardless of whether it is toggled directly or after switching OFF the PSLab MIC. This happens only if the PSLab Device isn't connected.

**Expected Behaviour**
The In-Built-MIC feature should always work, regardless of the fact that the PSLab MIC feature has been used first or not. 

**Steps to reproduce it**
Ensure that the PSLab Device isn't connected.
1. Click on the PSLab MIC checkbox to toggle it.
2. Either turn OFF the PSLab MIC first and then click on the In-Built-MIC checkbox, or click on the In-Built-MIC checkbox directly.
3. One will observe that the feature isn't working.

**Screenshots of the issue**
This is a screenshot of the OscilloscopeActivity after following the above steps:-
Screenshot 2024-05-19 184502
One can observe that there is no plot for the Built-In-MIC input.

Yes.

AsCress:
Fixes #2421 and #2173 
This issue was arising primarily due to the following two reasons:-

1. In the ChannelParametersFragment, whenever one checkbox was checked, the other was being unchecked without correcting the state of isAudioInputSelected, causing it to get toggled two times, resulting in it being false, even when one of the audio inputs was selected.
2. In the OscilloscopeActivity, the synchronized method was placed in the wrong scope, causing lock.wait() to get executed even when there was no PSLab Device connected. As there was no input, the thread would keep waiting forever, as lock.notify() would never be called. Hence, it would never detect a change in isAudioInputSelected.

## Changes 
- app/src/main/java/io/pslab/fragment/ChannelParametersFragment.java - corrected the state of isAudioInputSelected whenever a checkbox is checked, due to another getting unchecked.
- app/src/main/java/io/pslab/activity/OscilloscopeActivity.java - corrected the scope of the synchronized method, so that lock.wait() only gets executed if the PSLab Device is connected.

## Screenshots / Recordings  
This is a screenshot of the OscilloscopeActivity, after following the required steps to reproduce the issue, as mentioned in #2421.
WhatsApp Image 2024-05-19 at 6 29 45 PM
As one can see, the issue is resolved.

**Checklist**: 

### DIFF
--- a/app/src/main/java/io/pslab/fragment/ChannelParametersFragment.java
+++ b/app/src/main/java/io/pslab/fragment/ChannelParametersFragment.java
@@ -10,7 +10,6 @@ import androidx.annotation.NonNull;
 import androidx.core.content.ContextCompat;
 import androidx.fragment.app.Fragment;
 
-import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
--- a/app/src/main/java/io/pslab/fragment/ChannelParametersFragment.java
+++ b/app/src/main/java/io/pslab/fragment/ChannelParametersFragment.java
@@ -10,6 +10,7 @@ import androidx.annotation.NonNull;
 import androidx.core.content.ContextCompat;
 import androidx.fragment.app.Fragment;
 
+import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/pslab/fragment/ChannelParametersFragment.java
```diff
@@ -10,6 +10,7 @@
 import androidx.core.content.ContextCompat;
 import androidx.fragment.app.Fragment;
 
+import android.util.Log;
```
CloudyPadmal: Looks like this is unused?

AsCress: Oh yes! My bad, I had added these for debugging. I'm removing them.

AsCress:
@CloudyPadmal Done:)

marcnause:
I have tested the app without the fix and with the fix. I was able to reproduce the problem and I can confirm that I can toggle between both MICs now and the build-in microphone works whenever it is selected in the version which contains the fix.

marcnause:
@AsCress, just a general remark: I just noticed that you created this pull request on your development branch. I don't know if this was intentional or not. I like to (and sometimes forget to) make an extra branch for every issue I work on (named e.g. issue2421). That way I can switch between working on different issues. Even more convenient: Waiting for the merge of a pull request does not block me from working on other issues.

AsCress:
> @AsCress, just a general remark: I just noticed that you created this pull request on your development branch. I don't know if this was intentional or not. I like to (and sometimes forget to) make an extra branch for every issue I work on (named e.g. issue2421). That way I can switch between working on different issues. Even more convenient: Waiting for the merge of a pull request does not block me from working on other issues.

@marcnause Thank you very much for testing the fix and your comments!! I'll surely make extra branches for different issues from now on.

CloudyPadmal (approved):

PR #2430, opened 2024-05-25
fix: fixed the plots for the data captured through the In-Built-MIC

### LINKED ISSUE #2428: Wrong plots for the data captured through the In-Built MIC
This issue was also noticed by me in the pre-GSoC period, while writing the proposal.
The plot for the waveform captured using the In-Built MIC doesn't seem to be correct. This can be tested by using a tone generator.
**Actual Behaviour**
The captured waveform has different characteristics (time period and frequency) than the one generated by the tone generator.

**Expected Behaviour**
The captured waveform should have the same characteristics the one generated using the tone generator.

**Steps to reproduce it**
1. Turn on the In-Built MIC feature.
2. Use a tone generator to generate any kind of wave (say, sine) of a particular frequency.
3. You will observe that the one shown in the plot has a very different frequency.

**Screenshots of the issue**
Here, is the output for a sinusoidal waveform of frequency 1000 Hz generated using a tone generator.
WhatsApp Image 2024-05-23 at 4 36 19 PM
The output waveform has a frequency of approx. 20-25 Hz, which is a huge difference from the input waveform.

Yes.

AsCress:
Fixes #2428 
This issue was primarily arising due to two reasons:
**1. No manipulation of the xData before plotting -** The xData was not being manipulated before plotting in the case of the In-Built-MIC as it was being done in the other channels. As a result, the data sampled at 44,100 Hz (sampling period approx. 2.26 microseconds) was being plotted at intervals of 1 microsecond, which was causing a waveform of the wrong frequency to be displayed.

**2. Wrong configuration of the millisecond scales -** The xData was not being changed in the case of the In-Built-MIC when the user shifted to the millisecond scales using the seekBar in the TimebaseTriggerFragment, as it was being done in the case of the other channels. The xData was not being divided by 1000 when the millisecond scales were selected.

## Changes 
**- app/src/main/java/io/pslab/activity/OscilloscopeActivity.java -** Added the required data manipulations before plotting it.
**- app/src/main/java/io/pslab/others/AudioJack.java -** Made SAMPLING_RATE public to access it inside OscilloscopeActivity.

## Screenshots / Recordings 
The following is a screenshot for an audio wave of 1000Hz generated by a tone generator, captured by the In-Built-MIC:
Screenshot_20240525_082500
As one can see, the plot is now fixed.

**Checklist**: 

### DIFF
--- a/app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
+++ b/app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
@@ -1,6 +1,7 @@
 package io.pslab.activity;
 
 
+import static io.pslab.others.AudioJack.SAMPLING_RATE;
 import static io.pslab.others.MathUtils.map;
 
 import android.annotation.SuppressLint;
@@ -1027,12 +1028,14 @@ public class OscilloscopeActivity extends GuideActivity implements View.OnClickL
                         xDataString = new String[n];
                     }
                     for (int i = 0; i < n; i++) {
+                        float j = (float) (((double) i /SAMPLING_RATE)*1000000.0);
+                        j = j / ((timebase == 875) ? 1:1000);
                         float audioValue = (float) map(buffer[i], -32768, 32767, -3, 3);
                         if (!isFourierTransformSelected) {
                             if (noOfChannels == 1) {
-                                xDataString[i] = String.valueOf(2.0 * i);
+                                xDataString[i] = String.valueOf(j);
                             }
-                            entries.get(entries.size() - 1).add(new Entry(i, audioValue));
+                            entries.get(entries.size() - 1).add(new Entry(j, audioValue));
                         } else {
                             if (i < n / 2) {
                                 float y = (float) fftOut[i].abs() / samples;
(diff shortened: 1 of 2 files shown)


AsCress:
@marcnause Would you be able to confirm this fix, please ?

marcnause:
I tried with 2 phones again (1 as wave generator, 1 oscilloscope) and I got the following result:

image

Please notice that I chose a different timebase than in the screenshot in the first post. I notices two things: The waveform is cut off after 40 ms and at approximately 20 ms there is a point in the otherwise round wave form.

edit: It looks like the waveform is mirrored shortly after 20 ms. Is that possible?

AsCress:
@marcnause Thanks so much for testing the fix!! The issue which you mentioned about the cutting of the waveform is bound to happen since  we have corrected the scales but not changed the minBufferSize of the AudioJack reader. 
I'll be opening another issue ticket to tackle this since this is a very complicated issue and many solutions are possible, so I want to propose and discuss several solutions with you guys.
You will find that on low values of the timebase and waveform is displayed perfectly fine, as I have shown in the screenshot.
Can you confirm that the waveform being generated has the correct frequency, because that's the main issue that I want to tackle in this ticket, correcting the frequency of the waveform first.

marcnause:
@AsCress Everything you wrote makes sense! 👍 I can confirm that the waveform has the correct frequency as far as I can tell from looking at it.

mariobehling (approved):

PR #2465, opened 2024-06-04
fix: fixed the graph to clear whenever no channel is active

### LINKED ISSUE #2460: Oscilloscope: Graph doesn't clear after deselecting the In-Built MIC
**Actual Behaviour**
If one toggles the In-Built MIC ON by clicking on the radio button, and then turns it OFF the same way, the plot of the waveform still remains on the graph.

**Expected Behaviour**
Ideally, the graph should be cleared whenever a channel is turned OFF.

**Steps to reproduce it**

- Turn the In-Built MIC feature ON using the radio button.
- Turn it OFF the same way.
- One would observe that the plot for the input through the MIC is still there on the graph.

**Screenshots of the issue**
The following is a screenshot of the Oscilloscope screen after following the aforementioned steps: -

Screenshot_20240603_163514

Yes

**Note**
I've found this issue with the In-Built MIC only.
@marcnause Could you please find out if it exists for all the other channels as well ?

AsCress:
Fixes #2460 
Clears the graph whenever all the channels are toggled OFF.
## Changes 
- app/src/main/java/io/pslab/activity/OscilloscopeActivity.java

## Screenshots / Recordings  

https://github.com/fossasia/pslab-android/assets/125425881/4db79832-fe22-410d-b7f9-89b0e4ff6727

**Note:-**
Again, this is tested only with the In-Built MIC as shown.
@marcnause Could you please test this with all the other channels and when more than one channels are active as well ?

**Checklist**: 

## Summary by Sourcery

This pull request fixes the issue where the graph was not clearing when no channels were active. It also refactors the channel selection logic to improve code readability and maintainability.

- **Bug Fixes**:
    - Cleared the graph whenever all channels are toggled off to prevent displaying outdated data.
- **Enhancements**:
    - Refactored the channel selection logic to use a list for better readability and maintainability.

### DIFF
--- a/app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
+++ b/app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
@@ -511,6 +511,10 @@ public class OscilloscopeActivity extends GuideActivity implements View.OnClickL
                             audioJack.release();
                             audioJack = null;
                         }
+
+                        if (!(((isCH1Selected || isCH2Selected || isCH3Selected || isMICSelected) && scienceLab.isConnected()) || isInBuiltMicSelected)) {
+                            mChart.clearValues();
+                        }
                     }
                 }
             }
--- a/app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
+++ b/app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
@@ -341,6 +341,8 @@ public class OscilloscopeActivity extends GuideActivity implements View.OnClickL
 
         final Runnable runnable = new Runnable() {
 
+            private final List<String> channels = new ArrayList<>();
+
             @Override
             public void run() {
                 //Thread to check which checkbox is enabled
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
```diff
@@ -460,6 +460,10 @@ public void run() {
                             audioJack.release();
                             audioJack = null;
                         }
+
+                        if (!(((isCH1Selected || isCH2Selected || isCH3Selected || isMICSelected) && scienceLab.isConnected()) || isInBuiltMicSelected)) {
+                            mChart.clearValues();
```
marcnause: I get the following exception if I test the changes on my Motorola Moto X Play test device:

`FATAL EXCEPTION: Thread-4
Process: io.pslab, PID: 18636
android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
	at android.view.ViewRootImpl.checkThread(ViewRootImpl.java:6962)
	at android.view.ViewRootImpl.invalidateChildInParent(ViewRootImpl.java:1101)
	at android.view.ViewGroup.invalidateChild(ViewGroup.java:5209)
	at android.view.View.invalidateInternal(View.java:13659)
	at android.view.View.invalidate(View.java:13623)
	at android.view.View.invalidate(View.java:13607)
	at com.github.mikephil.charting.charts.Chart.clearValues(Chart.java:324)
	at io.pslab.activity.OscilloscopeActivity$1.run(OscilloscopeActivity.java:465)
	at java.lang.Thread.run(Thread.java:761)
`

marcnause: I changed line 465 to

` mChart.post(() -> mChart.clearValues());`

to run the clear command on the UI thread. The app does not crash anymore after that, but it becomes unresponsive. I think the reason is that the chart is cleared over and over again in an endless loop in the Runnable which runs in the extra thread.

I think it needs to be ensured that the chart is only cleared if something has been added to it before. I have seen that the Chart has an `isEmpty()` method which I think could be used for this check.

AsCress: @marcnause Thanks so much for testing. I understood the error, I am not sure why it comes up on some devices because on the device I tested, the solution worked fine. In any case, I guess I have a better solution in mind which would eliminate all this hassle of the _clearValues()_ function being called again and again from a thread which isn't the UI one. We could perhaps migrate refreshing of the graph to the fragment for the workaround. I'll make a commit and update this PR :))

marcnause: I think it makes a difference if the PSLab device is connected or not. Looking forward to your PR!

AsCress: @marcnause Yes that is evident. I'll soon be making a commit. I'll take care of that.

AsCress:
@marcnause Tested your fix on a couple of devices. Seems fine to me. I've added the _isEmpty()_ check which solves the issue I guess. Can you please check and confirm ?

marcnause:
@AsCress I tested with your latest changes and the UI was responding at all times. The graph was cleared when CH1, CH2, and CH3 were selected and the deselected. 

The only case where I had problems was when the PSLab device was connected and one of the MIC options was used. Then the graph was not cleared anymore when all channels were disabled.

I suspect that the boolean expression which determines that the graph is to be cleared is not quite correct, but I don't see the problem at the moment.

AsCress:
@marcnause Ok, got that. I'll go through the whole thing again and update this PR today.

marcnause:
@AsCress Your code was perfectly fine, the problem was located somewhere else in the OscilloscopeActivity. I have made some changes in https://github.com/fossasia/pslab-android/pull/2465/commits/a00fb2c2a59a7f14bc9b33190b898f631232b725 and I think the problem I reported previously should be fixed now. 

Can you please review my changes?

marcnause (approved):

app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
```diff
@@ -350,154 +352,35 @@ public void run() {
...
-                        if (scienceLab.isConnected() && isCH1Selected && isCH2Selected && isCH3Selected && !isAudioInputSelected && !isXYPlotSelected) {
-                            captureTask = new CaptureTask();
-                            captureTask.execute(CHANNEL.CH1.toString(), CHANNEL.CH2.toString(), CHANNEL.CH3.toString());
-                            synchronized (lock) {
-                                try {
-                                    lock.wait();
-                                } catch (InterruptedException e) {
-                                    e.printStackTrace();
-                                }
+                            if (isAudioInputSelected && isInBuiltMicSelected || (scienceLab.isConnected() && isMICSelected)) {
```
AsCress, line 379: @marcnause We may have to improve this condition/add more conditions for handling the In-Built MIC with this logic. If the _PSLab_ device is connected and the In-Built MIC feature of the phone is toggled _ON_ using the radio button, it won't get toggled _OFF_ at all by unchecking the radio button, and would keep running forever. Could you please once test and confirm if this issue exists ?

marcnause: You are right, the issue exists!

mariobehling:
@sourcery-ai review

app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
marcnause: @AsCress I think this fix was fixed (😉) by the changes you introduced in https://github.com/fossasia/pslab-android/pull/2529. I  did some tests and I *think* it works as intended now. Can you confirm?

AsCress: @marcnause Yes you are correct :))

PR #2474, opened 2024-06-11
feat: re-activate color overlay for pin layout

### LINKED ISSUE #2467: Use hidden color map in pin layout screens
**Actual Behaviour**

The pin layout screens load an image which contains color coding of the pins, but it is never visible since it is set to "invisible" in `fragment_pin_layout.xml`. 

Replacing `android:visibility="invisible"` with `android:alpha="0.5"` in `fragment_pin_layout.xml` makes the color map visible and lets the image of the PSLab shine through.

It is unclear why the image is hidden.

**Expected Behaviour**

The color map is used and the colors correspond with the colors in the popups.

**Steps to reproduce it**

1. Open app.
2. Tap menu symbol.
3. Choose front layout or back layout.

**Screenshots of the issue**
***Current layout***
Screenshot_20240604-202850

***Changed layout with added alpha value***
Screenshot_20240604-201115

***Popup with color which does not match color in color map***
Screenshot_20240604-201128

Yes.

marcnause:
Fixes #2467

## Changes 
- made color overlay visible again
- moved hard-coded strings and colors to resources
- fixed typo for TXD and RXD (was TXP and RXP)
- close pop-up with single tap (instead of 2 taps which were required before)
- prevent possible NullPointerException when pop-up is created
- fixed a few lint warnings
- force garbage collection for free memory more aggressively to avoid OurOfMemoryErrors on older devices

## Screenshots / Recordings  

https://github.com/fossasia/pslab-android/assets/11596220/20705294-89a5-479a-9c86-7475cd2d8ef8

**Checklist**: 

### DIFF
--- a/app/src/main/java/io/pslab/fragment/PSLabPinLayoutFragment.java
+++ b/app/src/main/java/io/pslab/fragment/PSLabPinLayoutFragment.java
@@ -1,6 +1,8 @@
 package io.pslab.fragment;
 
 import android.annotation.SuppressLint;
+import android.app.Activity;
+import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.graphics.Matrix;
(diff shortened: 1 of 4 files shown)


CloudyPadmal:
This looks good. I wonder if we should update the image of the board in this PR or in a different one. As this shows the pinout for an older version of the board.

CloudyPadmal (approved):

marcnause:
I think we should definitely update the pinout or provide a version for v5 and one for v6 (I am unsure which variant is preferable), but I would like to do that in a second PR to keep the scope of this PR small.

CloudyPadmal:
Sounds good 👍

Issue #2101, opened 2020-02-03
Oscilloscope: Trigger CH1 actually works on CH2


mariobehling:
Oscilloscope: Trigger CH1 actually works on CH2.

CodeWithSouma:
@mariobehling  sir, please help me how to fix this problem I want to work on this issue.

PrajwalG22:
Hey I have just started with open source contribution.
Can you help me working on this issue.

suman-somu:
working on this. 
good to me, if assigned.

thank you.

MonuGit9:
Claim

marcnause:
I close this issue since triggering works as intended after https://github.com/fossasia/pslab-android/pull/2482 has been merged.

PR #2482, opened 2024-06-15
feat: added Triggering Controls

### LINKED ISSUE #2476: Oscilloscope: Add new feature for Triggering
This is for the addition of the following proposed feature:-

**Triggering Controls**:
- Currently, the oscilloscope supports only a single type of edge – triggering.
- All three-types of edge triggering: Rising-edge, Falling-edge and Dual-edge can be added.

**Actual Behaviour**
Currently, we only support hardware based triggering, i.e., using the trigger controls built in the PSLab device. This also has limitations as certain channels support only a single kind of triggering.

**Expected Behaviour**
Triggering is a feature which we can implement at a software level in the Android app. This would ensure that all kinds of triggering controls are available for all the channels.

**Steps to reproduce it**
N/A

**LogCat for the issue**
N/A

**Screenshots of the issue**
N/A

Yes

AsCress:
Resolves #2476 
Implements _Triggering Controls_ at a **software** level within the application.

**Triggering Controls**:
- Currently, the oscilloscope supports only a single type of edge – triggering.
- All three-types of edge triggering: Rising-edge, Falling-edge and Dual-edge can
be added.

The UI for the feature is implemented in the form of a _Spinner_ in the _TimebaseTriggerFragment_.
## Changes 
- Currently, the app only supports **Hardware Triggering** using the in-built feature in the PSLab Device. There is no triggering feature for the In-Built MIC (of the Android phone) as well.
- The Hardware Triggering feature has a lot of bugs, channel limitations, and doesn't work (refer #2101).
- This PR removes the current Hardware Triggering feature from the app and instead implements it at a **software** level, using _Single_ sweep mode, and an algorithm which works much more reliably, allowing the user to have access to all three types of edge triggering (Rising, Falling, Dual) for all the channels (CH1, CH2, CH3, PSLab MIC and In-Built MIC).

## Screenshots / Recordings  
Here is a demonstration of the newly added feature using **CH1** - 

https://github.com/fossasia/pslab-android/assets/125425881/621f8e23-25fb-4e81-ba64-8f156dd52f0c

**Checklist**: 

@marcnause @CloudyPadmal @bessman How do you find this new implementation of _Triggering_ in the app ?
We can also possibly enhance this feature by adding text inputs for the trigger value, as done for the Position Controls feature. How does that seem to you ?

### DIFF
--- a/app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
+++ b/app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
@@ -122,6 +122,7 @@ public class OscilloscopeActivity extends GuideActivity implements View.OnClickL
     public static boolean isInBuiltMicSelected;
     public boolean isAudioInputSelected;
     public boolean isTriggerSelected;
+    public boolean isTriggered;
     public boolean isFourierTransformSelected;
     public boolean isXYPlotSelected;
     private boolean isDataAnalysisFragSelected;
@@ -130,6 +131,7 @@ public class OscilloscopeActivity extends GuideActivity implements View.OnClickL
     public boolean isCH1FrequencyRequired;
     public boolean isCH2FrequencyRequired;
     public String triggerChannel;
+    public String triggerMode;
     public String curveFittingChannel1;
     public String curveFittingChannel2;
     public String xyPlotXAxisChannel;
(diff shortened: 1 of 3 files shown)


marcnause:
I tested the feature and it worked really well for me. 👍

If it is no big effort, it would be nice if you could add text inputs.

I am not sure about the effort an d if it really is as useful as I think it is, but I have another proposal for an optimization: I think it would be easier to set the trigger value (especially for waves with small amplitudes) if the range of the `SeekBar` would auto scale in a way similar to what you implemented in https://github.com/fossasia/pslab-android/pull/2471.

AsCress:
@marcnause Yeah ! You're right. I was gonna suggest the same thing. I'll implement all these features and update this PR today.

AsCress:
@marcnause I've added the above mentioned features. Can you please test these ?

marcnause (approved):

PR #2488, opened 2024-06-17
feat: implemented serial driver to facilitate communication with PSLab Devices

### LINKED ISSUE #2472: PSLab v6 can't communicate with the Android device
**Actual Behaviour**
Due to the PSLab v6 device using a new Serial Bridge Controller (the CP2102N), the _CommunicationHandler.java_ currently doesn't support communication with the device.

**Expected Behaviour**
We should update the class to support serial communication with the PSLab v6 device (using API AN571), alongwith supporting backwards compatibility for the v5.

**Steps to reproduce it**
N/A

**LogCat for the issue**
N/A

**Screenshots of the issue**
N/A

@bessman @CloudyPadmal This is the issue ticket for coordinating on the progress.

AsCress:
Resolves #2472 
Implements the [usb-serial-for-android](https://github.com/mik3y/usb-serial-for-android) library to facilitate communication with the _PSLab_ device, as discussed in our weekly GSoC meet.

## Changes 
- This PR reconfigures the _CommunicationHandler.java_ class to implement the [usb-serial-for-android](https://github.com/mik3y/usb-serial-for-android) library, in order to facilitate communication with the new V6 device, while also having backwards compatibility with the V5.

## Screenshots / Recordings  
N/A

**Checklist**: 

**Note: -**
I have only been able to test this new driver with the V6 device.
@marcnause Could you please help me with testing this on the V5 device and confirm if all the features work fine ?

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -76,6 +76,7 @@ dependencies {
     implementation("de.hdodenhof:circleimageview:3.1.0")
     implementation("com.github.devlight.navigationtabstrip:navigationtabstrip:1.0.4")
     implementation("com.afollestad.material-dialogs", "commons", "0.9.6.0")
+    implementation("com.github.mik3y:usb-serial-for-android:3.7.3")
     implementation("com.github.medyo:android-about-page:1.3.1")
     implementation("com.github.tiagohm.MarkdownView:library:0.19.0")
     implementation("com.github.mirrajabi:search-dialog:1.2.4")
--- a/app/src/main/res/xml/device_filter.xml
+++ b/app/src/main/res/xml/device_filter.xml
@@ -26,6 +26,7 @@
     <usb-device vendor-id="6790" product-id="29987" /> <!-- 0x7523: CH340 -->
 
     <!-- CDC driver -->
+    <usb-device vendor-id="1240" product-id="223"/> <!-- 1240 / 223: MCP2200(For PSLab V5) -->
     <usb-device vendor-id="9025" />                   <!-- 0x2341 / ......: Arduino -->
     <usb-device vendor-id="5824" product-id="1155" /> <!-- 0x16C0 / 0x0483: Teensyduino  -->
     <usb-device vendor-id="1003" product-id="8260" /> <!-- 0x03EB / 0x2044: Atmel Lufa -->
--- a/app/src/main/res/xml/device_filter.xml
+++ b/app/src/main/res/xml/device_filter.xml
@@ -1,26 +1,38 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <!-- For PSLab -->
-    <usb-device vendor-id="1240" product-id="223" />
+    <!-- 0x0403 / 0x60??: FTDI -->
+    <usb-device vendor-id="1027" product-id="24577" /> <!-- 0x6001: FT232R -->
+    <usb-device vendor-id="1027" product-id="24592" /> <!-- 0x6010: FT2232H -->
+    <usb-device vendor-id="1027" product-id="24593" /> <!-- 0x6011: FT4232H -->
+    <usb-device vendor-id="1027" product-id="24596" /> <!-- 0x6014: FT232H -->
+    <usb-device vendor-id="1027" product-id="24597" /> <!-- 0x6015: FT230X, FT231X, FT234XD -->
 
-    <!-- 0x0403 / 0x6001: FTDI FT232R UART -->
-    <usb-device vendor-id="1027" product-id="24577" />
+    <!-- 0x10C4 / 0xEA??: Silabs CP210x -->
+    <usb-device vendor-id="4292" product-id="60000" /> <!-- 0xea60: CP2102 and other CP210x single port devices -->
+    <usb-device vendor-id="4292" product-id="60016" /> <!-- 0xea70: CP2105 -->
+    <usb-device vendor-id="4292" product-id="60017" /> <!-- 0xea71: CP2108 -->
 
-    <!-- 0x0403 / 0x6015: FTDI FT231X -->
-    <usb-device vendor-id="1027" product-id="24597" />
+    <!-- 0x067B / 0x23?3: Prolific PL2303x -->
+    <usb-device vendor-id="1659" product-id="8963" /> <!-- 0x2303: PL2303HX, HXD, TA, ... -->
+    <usb-device vendor-id="1659" product-id="9123" /> <!-- 0x23a3: PL2303GC -->
+    <usb-device vendor-id="1659" product-id="9139" /> <!-- 0x23b3: PL2303GB -->
+    <usb-device vendor-id="1659" product-id="9155" /> <!-- 0x23c3: PL2303GT -->
+    <usb-device vendor-id="1659" product-id="9171" /> <!-- 0x23d3: PL2303GL -->
+    <usb-device vendor-id="1659" product-id="9187" /> <!-- 0x23e3: PL2303GE -->
+    <usb-device vendor-id="1659" product-id="9203" /> <!-- 0x23f3: PL2303GS -->
 
-    <!-- 0x2341 / Arduino -->
-    <usb-device vendor-id="9025" />
+    <!-- 0x1a86 / 0x?523: Qinheng CH34x -->
+    <usb-device vendor-id="6790" product-id="21795" /> <!-- 0x5523: CH341A -->
+    <usb-device vendor-id="6790" product-id="29987" /> <!-- 0x7523: CH340 -->
 
-    <!-- 0x16C0 / 0x0483: Teensyduino  -->
-    <usb-device vendor-id="5824" product-id="1155" />
-
-    <!-- 0x10C4 / 0xEA60: CP210x UART Bridge -->
-    <usb-device vendor-id="4292" product-id="60000" />
-
-    <!-- 0x067B / 0x2303: Prolific PL2303 -->
-    <usb-device vendor-id="1659" product-id="8963" />
-
-    <!-- 0x1a86 / 0x7523: Qinheng CH340 -->
-    <usb-device vendor-id="6790" product-id="29987" />
+    <!-- CDC driver -->
+    <usb-device vendor-id="9025" />                   <!-- 0x2341 / ......: Arduino -->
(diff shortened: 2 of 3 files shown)


app/src/main/res/xml/device_filter.xml
```diff
@@ -1,26 +1,38 @@
...
-
-    <!-- 0x067B / 0x2303: Prolific PL2303 -->
-    <usb-device vendor-id="1659" product-id="8963" />
-
-    <!-- 0x1a86 / 0x7523: Qinheng CH340 -->
-    <usb-device vendor-id="6790" product-id="29987" />
+    <!-- CDC driver -->
+    <usb-device vendor-id="9025" />                   <!-- 0x2341 / ......: Arduino -->
+    <usb-device vendor-id="5824" product-id="1155" /> <!-- 0x16C0 / 0x0483: Teensyduino  -->
+    <usb-device vendor-id="1003" product-id="8260" /> <!-- 0x03EB / 0x2044: Atmel Lufa -->
```
CloudyPadmal, line 32: Do we need all these other IDs since we will never use those?

AsCress: @CloudyPadmal You are right. We never use these. However, we already had a list of such devices in our project, I just appended it to make it more complete. The serial library can use this list to find any of these devices if at all we implement them in the future. Does it sound right or shall we remove these ?

marcnause:
@AsCress I tested with the APK I downloaded via the link that was provided by the pipeline. Thumbs up for PR https://github.com/fossasia/pslab-android/pull/2487! 😉

I tested the app on my Motorola Moto X Play and the PSLab v5 device.

During testing I used the Wave Generator, the Oscilloscope, the Power Supply and the Multimeter which worked in general, but I noticed a few issues.

As expected, the app recognizes the PSLab v5 device when it is connected to the phone before the app is started.

The phone will ask if I want to allow access to the device and if I want the app to open every time the PSLab v5 device is connected. The settings seem to be not persisted, I get asked the same question on every startup.

The app notices when the PSLab v5 device is disconnected, but it does not recognize if the the PSLab v5 device is connected again.

The Oscilloscope shows glitches in the waveforms for sine waves and digital waves created with the Wave Generator. The glitches don't show up all of the time, but they are very noticeable.

Screenshot_20240617-203841
Screenshot_20240617-204131

Please let me know if there is anything I can do to help you debugging these issues.

app/src/main/res/xml/device_filter.xml
CloudyPadmal: I guess it does not hurt to have it just listed.

CloudyPadmal (approved):
Looks good!

AsCress:
@marcnause Thank you very much for testing on the V5 device.
From all the information that you provided and by observing the waveforms, the best issue I can guess here is that the device is having communication problems with the app (since I was facing similar issues with the V6 initially), as we are using the CDC ACM driver for the V5 device, and I couldn't test it, since I don't have the V5 device.
As far as the debugging is concerned, it would be great if you could provide me the _logcat_ files generated while running the app when you encounter these issues.
It took me a couple of hours to debug the V6 device and make it error free, but it was possible only because I had that one with me.
I'll try my best to do whatever I can by just looking at the logs. 🙂

AsCress:
@marcnause I've fixed the permission issues. Should be retaining permissions now. Can you please confirm that as well ?

marcnause:
@AsCress Retaining permission works now. When I connect the PSLab v5 to my phone the app opens. The device is also recognized after re connection now.

I noticed one problem though. If I wait approximately 10 seconds until I disconnect the PSLab v5, everything works fine. If I wait only a few seconds (I counted from 1 to 5), the app crashes when I disconnect the device.

Here is a log file where I connect he device, wait for approximately 10 seconds, re-connect the device and then remove it after approximately 5 seconds.

[crash.txt](https://github.com/user-attachments/files/15892126/crash.txt)

I also created a sine wave in the Wave Generator, but now there is no waveform visible in the Oscilloscope. I have captured the log and it is visible that the data of teh wave contains only zeros.
 
[pslab_sine.txt](https://github.com/user-attachments/files/15892234/pslab_sine.txt)

I did the same with a digital wave and the waveform is a horizontal line with spikes instead of squares.

[pslab_digital.txt](https://github.com/user-attachments/files/15892249/pslab_digital.txt)

Screenshot_20240618-222332

AsCress:
@marcnause Thank you very much for testing and the logs. They helped me very much. I've added some instructions in there to start communication with the v5 device. Can you please test if the communication works now ?

marcnause:
@AsCress I was able to crash the app by disconnecting after a few seconds, but I had to find the sweet spot.

[crash.txt](https://github.com/user-attachments/files/15903780/crash.txt)

I still get weird waveforms in the Oscilloscope. I created a sine wave with the Wave Generator and got this result. 🤔

[sine.txt](https://github.com/user-attachments/files/15903710/sine.txt)

Screenshot_20240619-180915

If there is anything in particular I can do to support you, please let me know.

AsCress:
@marcnause Just trying a quick fix to check if this fixes the issues. Could please test this again ? Thanks.

marcnause:
@AsCress I tested the latest version, but as far as I can tell, the behavior visible to the user did not change.

What I did:

1. Open app
2. Connect PSLab v5
3. Set frequency for sine wave for CH1 in Wave Generator
4. Start Oscilloscope from Wave Generator
5. Choose CH1 in Oscilloscope
6. Look at waveform
7. Disconnect PSLab v5
8. Tap back button until app closes

Here is the log:

[log.txt](https://github.com/user-attachments/files/15905422/log.txt)

AsCress:
@marcnause Thanks for the testing. I've just pushed in a new version. Tested this today alongwith @bessman using an old android phone and a v5 device, remotely. The communication worked (there were no read errors as such in the logs) however, there were spikes as you had mentioned before and the amplitude of noise on any unconnected channel was too high. Can you please once test this version with your setup and check if the same issue persists ? It'd be great if you could provide logs as well. Thanks !!

**Edit:-**
The latest version I'm pushing now should solve the crash issue as well.

marcnause:
I still see errors in the log of my phone: [log.txt](https://github.com/user-attachments/files/15920045/log.txt)

And I was still able to provoke crashes: [crash.txt](https://github.com/user-attachments/files/15920060/crash.txt)

I just remembered that @bessman mentioned that there are at least 2 versions of the firmware for the PSLab v5. I got my device in 2019 and never updated the firmware. I don't know if it makes any difference.

bessman:
> there are at least 2 versions of the firmware for the PSLab v5.

I don't think I've said that. As far as I know all v5 boards shipped with firmware version 2.0. Perhaps you misheard "firmware version 2" as "two firmware versions"?

marcnause:
> Perhaps you misheard "firmware version 2" as "two firmware versions"?

Oh, OK! That was a misunderstanding on my part.

AsCress:
@marcnause @bessman This is quite strange though 🤔🤔. An implementation which works on one V5 device doesn't work at all on another. Any inputs as to what could be causing it ? Nonetheless, I plan to take on this topic in detail in our weekly meet.

marcnause:
So far I only tested with one phone. Maybe it would be a good idea to try other devices to rule out an issue with the phone. I will do that later today and post the results here.

AsCress:
@marcnause That would be great !!

marcnause:
@AsCress I tested again with different Android devices, I downloaded the APK from Github again and also built an APK from your branch, but always with the same result described above already.

Do you still have any uncommitted changes? Do you still have the APK you used for testing with the PSLab v5 (just to rule out any mistakes on my side)? Maybe we can discuss the setup you used for testing in our weekly meeting. Is it possible that it had an impact on the communication (e.g. timing)?

marcnause:
I tested again today with different cables and different versions of the app. Turns out that the problem was the way I had connected outputs and inputs of my PSLab. I had connected PV1 to VOL and there is a bug which prevents the board to boot properly if this connection is present during startup (https://github.com/fossasia/pslab-firmware/issues/142). The stupid thing is, that I reported the bug myself last year. 🤦 @AsCress, I have to apologize to you for wasting your time.

Once I removed the cable, I was able to observe a waveform in the oscilloscope (with occasional glitches as described in https://github.com/fossasia/pslab-android/pull/2488#issuecomment-2174214366).

I noticed the "PV1/VOL-bug" when I used the PSLab Python library. The Python library reports "Device not found". Would it be possible to detect that the PSLab is in the weird state in the Android app and warn the user about the bug? If yes, we should create another ticket for that.

AsCress:
@marcnause Not an issue at all !! This gave me an opportunity to research different implementations of the CDC ACM driver 😂. We now have a direction along which we can work to solve the issue ! I've just pushed a new version. Could you test if this fixes those _oscassional glitches_ in the waveform ?

marcnause:
I tested with the latest version and I got an IndexOutOfBoundsExeption in the line you changed in https://github.com/fossasia/pslab-android/pull/2488/commits/84a7dceb3cedc4076ab1f748f59f2fecf5384415. 

Turns out, that interface has only one endpoint on my test device. Changing the the index of `getEndpoint` to `0` fixed the crash, but then I did not receive any data. I dug a little bit deeper and I found out that my device seems to have 3 USB interfaces, even though it has only one USB jack. I have yet to find the combination of interface- and endpoint-index that works for my test device.

marcnause:
I tested with the right indices for my device (1 and 1). I got a correct waveform I did not see any glitches anymore. 👍

AsCress:
@marcnause Thanks for confirming that it works fine. I've pushed a final new version which integrates this fix into our code. This PR is now ready to be merged if everything works fine.

app/build.gradle.kts
```diff
@@ -76,6 +76,7 @@ dependencies {
     implementation("de.hdodenhof:circleimageview:3.1.0")
     implementation("com.github.devlight.navigationtabstrip:navigationtabstrip:1.0.4")
     implementation("com.afollestad.material-dialogs", "commons", "0.9.6.0")
+    implementation("com.github.mik3y:usb-serial-for-android:3.7.3")
```
marcnause, line 79: Do we need to add information that we use this library (and the license) somewhere in the app?  

Should we create an extra issue for that? I don't think there is a screen which lists all used libraries with their licenses.

marcnause: I added a new issue: https://github.com/fossasia/pslab-android/issues/2498

marcnause (approved):

PR #2489, opened 2024-06-17
Improve PowerSourceActivity

marcnause:
## Changes 
- Don't lose values on rotate if no PSLab device is connected.
- Moveed upper and lower bounds of power source channels to extra data structure (removes many numbers from the code).
- Respect locale in `EditText`s (Methods of Apache Commons NumberUtils don't use the locale)
- Tried to simplify code.

**Checklist**: 

### DIFF
--- a/app/src/main/java/io/pslab/activity/PowerSourceActivity.java
+++ b/app/src/main/java/io/pslab/activity/PowerSourceActivity.java
@@ -213,7 +213,7 @@ public class PowerSourceActivity extends GuideActivity {
             public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
                 if (actionId == EditorInfo.IME_ACTION_DONE) {
                     final String voltageValue = remove(displayPV1.getText(), "V", "\\+").trim();
-                    final float voltage = PV1_VOLTAGE_RANGE.clamp(parseFloat(voltageValue, 0.0f));
+                    final float voltage = PV1_VOLTAGE_RANGE.clamp(parseFloat(voltageValue, PV1_VOLTAGE_RANGE.getLower()));
                     setText(displayPV1, VOLTAGE_FORMAT, voltage);
                     controllerPV1.setProgress(mapPowerToProgress(voltage, PV1_CONTROLLER_MAX,
                             PV1_VOLTAGE_RANGE.getUpper(), PV1_VOLTAGE_RANGE.getLower()));
@@ -227,7 +227,7 @@ public class PowerSourceActivity extends GuideActivity {
             public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
                 if (actionId == EditorInfo.IME_ACTION_DONE) {
                     final String voltageValue = remove(displayPV2.getText(), "V", "\\+").trim();
-                    final float voltage = PV2_VOLTAGE_RANGE.clamp(parseFloat(voltageValue, 0.0f));
+                    final float voltage = PV2_VOLTAGE_RANGE.clamp(parseFloat(voltageValue, PV2_VOLTAGE_RANGE.getLower()));
                     setText(displayPV2, VOLTAGE_FORMAT, voltage);
                     controllerPV2.setProgress(mapPowerToProgress(voltage, PV2_CONTROLLER_MAX,
                             PV2_VOLTAGE_RANGE.getUpper(), PV2_VOLTAGE_RANGE.getLower()));
@@ -241,7 +241,7 @@ public class PowerSourceActivity extends GuideActivity {
             public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
                 if (actionId == EditorInfo.IME_ACTION_DONE) {
                     final String voltageValue = remove(displayPV3.getText(), "V", "\\+").trim();
-                    final float voltage = PV3_VOLTAGE_RANGE.clamp(parseFloat(voltageValue, 0.0f));
+                    final float voltage = PV3_VOLTAGE_RANGE.clamp(parseFloat(voltageValue, PV3_VOLTAGE_RANGE.getLower()));
                     setText(displayPV3, VOLTAGE_FORMAT, voltage);
                     controllerPV3.setProgress(mapPowerToProgress(voltage, PV3_CONTROLLER_MAX,
                             PV3_VOLTAGE_RANGE.getUpper(), PV3_VOLTAGE_RANGE.getLower()));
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/pslab/activity/PowerSourceActivity.java
```diff
@@ -29,8 +33,8 @@
 import com.sdsmdg.harjot.crollerTest.Croller;
 import com.sdsmdg.harjot.crollerTest.OnCrollerChangeListener;
 
-import org.apache.commons.lang3.math.NumberUtils;
```
CloudyPadmal, line 32: Didn't @AsCress work on this few days ago? 🤔

marcnause: Yes, he did! The method  NumberUtils.isCreatable(String) from the library he used did a good job preventing the crash in the Activity, but I noticed that it does not take the locale into account which lead to numbers not being parsed on my test device. This caused the values to be reset on orientation change when the PSLab is not connected which I did not notice until I added the onSaveInstanceState method.

AsCress: @marcnause @CloudyPadmal Yes, this is a perfectly right approach to avoid locales not getting parsed. However, I just had one or points to mention about this. I'll just send these in a review.

app/src/main/java/io/pslab/activity/PowerSourceActivity.java
```diff
@@ -190,26 +212,11 @@ public void onClick(View v) {
...
-                        controllerPV1.setProgress(mapPowerToProgress(voltage, PV1_CONTROLLER_MAX,
-                                5.00f, -5.00f));
-                    } else {
-                        float voltage = -5.00f;
-                        displayPV1.setText(voltage + " V");
-                        controllerPV1.setProgress(mapPowerToProgress(voltage, PV1_CONTROLLER_MAX,
-                                5.00f, -5.00f));
-                    }
+                    final String voltageValue = remove(displayPV1.getText(), "V", "\\+").trim();
+                    final float voltage = PV1_VOLTAGE_RANGE.clamp(parseFloat(voltageValue, 0.0f));
```
AsCress: Perhaps we could reset PV1 to its actual default value (-5.00 V) instead of 0.00 V as well, as we are doing for all the other controllers. Does that sound right ?

app/src/main/java/io/pslab/activity/PowerSourceActivity.java
```diff
@@ -219,26 +226,11 @@ public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
...
-                        controllerPV2.setProgress(mapPowerToProgress(voltage, PV2_CONTROLLER_MAX,
-                                3.30f, -3.30f));
-                    } else {
-                        float voltage = -3.30f;
-                        displayPV2.setText(voltage + " V");
-                        controllerPV2.setProgress(mapPowerToProgress(voltage, PV2_CONTROLLER_MAX,
-                                3.30f, -3.30f));
-                    }
+                    final String voltageValue = remove(displayPV2.getText(), "V", "\\+").trim();
+                    final float voltage = PV2_VOLTAGE_RANGE.clamp(parseFloat(voltageValue, 0.0f));
```
AsCress: The same for this one as well.

app/src/main/java/io/pslab/activity/PowerSourceActivity.java
```diff
@@ -248,26 +240,11 @@ public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
...
-                        }
-                        controllerPV3.setProgress(mapPowerToProgress(voltage, PV3_CONTROLLER_MAX,
-                                3.30f, 0.00f));
-                    } else {
-                        float voltage = 0.00f;
-                        displayPV3.setText(voltage + " V");
-                        controllerPV3.setProgress(mapPowerToProgress(voltage, PV3_CONTROLLER_MAX,
-                                3.30f, 0.00f));
-                    }
+                    final String voltageValue = remove(displayPV3.getText(), "V", "\\+").trim();
```
AsCress, line 243: So, currently we are not supporting negative numbers to be inputted via the _EditTexts_ (we have `android:inputType` set to `numberDecimal` for _displayPV1_ and _displayPV2_). 
However, those controllers can support negative voltages as well, and we should implement negative inputs. But when we do implement those, we'd be having many corner cases to handle so as to make sure that the user input is float-parseable. Wouldn't that increase the complexity of the operations we perform on the string before parsing it ? Or is it handleable using this approach ?
Please let me know if this concern sounds right.

AsCress (changes requested):

app/src/main/java/io/pslab/activity/PowerSourceActivity.java
CloudyPadmal: Excellent!

app/src/main/java/io/pslab/activity/PowerSourceActivity.java
marcnause: You are right, I have changed the default value and will push a new version.

app/src/main/java/io/pslab/activity/PowerSourceActivity.java
marcnause: See above.

app/src/main/java/io/pslab/activity/PowerSourceActivity.java
marcnause: I tried setting the `android:inputType` to `numberSigned|numberDecimal` which nearly did the trick, but unfortunately the keyboard of my test phone is more or less broken. The keyboard has point and comma, but the comma is disabled, even when using the German locale. I can only use the point, but then the String cannot be parsed.

If I edit the text carefully and don't delete the comma, I can set negative values without any changes in the code.

There are more or less clunky workarounds for this problem (see https://stackoverflow.com/questions/3821539/decimal-separator-comma-with-numberdecimal-inputtype-in-edittext/51983392), but I would propose that we keep the current behavior for now and I will create an extra issue for this problem.

edit: I created https://github.com/fossasia/pslab-android/issues/2492

AsCress: @marcnause Sure!

AsCress (approved):

CloudyPadmal (approved):

Issue #2490, opened 2024-06-18
Update commands to achieve compatibility with the PSLab v6 device


AsCress:
**Actual Behaviour**
In the [latest](https://github.com/fossasia/pslab-firmware/releases/tag/v3.0.0) firmware for the v6 device, we have changed many commands and removed other redundant ones. This makes the current command implementation in the _Android_ app outdated and hence many features (such as the _Oscilloscope_, _PowerSource_, etc.) arn't working.

**Expected Behaviour**
We would have to update the _Android_ app to make it a 1:1 port from the [python](https://github.com/fossasia/pslab-python) implementation so as to make it compatible with both PSLab v5 and v6.

**Steps to reproduce it**
N/A

**LogCat for the issue**
N/A

**Screenshots of the issue**
N/A

**Would you like to work on the issue?**
Yes

PR #2494, opened 2024-06-22
feat: added Automatic Measurements

### LINKED ISSUE #2491: Add new feature for Automated Measurements
This is for the addition of the following proposed feature:-

**Automated Measurements**:
- The oscilloscope as of now doesn’t have a feature to automatically measure and display characteristics of the waveform such as amplitude, frequency, period, etc.
- This feature can be added which would calculate the characteristics of the waveform using algorithms and display them.

**Actual Behaviour**
Currently, we have no feature in the app using which the characteristics (such as the _frequency_, _period_, _amplitude_, etc.) of the input waveforms can be measured.

**Expected Behaviour**
We can implement a feature using algorithms at the **software** level to measure all these characteristics within the app and display them to the user.

**Steps to reproduce it**
N/A

**LogCat for the issue**
N/A

**Screenshots of the issue**
N/A

Yes

AsCress:
Resolves #2491 
Implements _Automatic Measurements_ of waveform characteristics within the app.

**Automatic Measurements**:
- The oscilloscope as of now doesn’t have a feature to automatically measure and display characteristics of the waveform such as amplitude, frequency, period, etc.
- This feature can be added which would calculate the characteristics of the waveform using algorithms and display them.

The UI for the feature is implemented in the form of a _Checkable MenuItem_ in the _Menu_ of the _OscilloscopeActivity_. Checking it would activate the feature, showing the measurements in _TextViews_ populated inside a _RecyclerView_.

## Changes 
- The feature is added in such a way that it doesn't cause **crowding** of elements in the _OscilloscopeActivity_, and can be used whenever wanted by activating it.
- When activated, common waveform characteristics such as _frequency_, _period_, _amplitude_, etc. of the waveforms selected are calculated using **software** algorithms within the app and the data is used to populate the _RecyclerView_ after every successful capture.
- In order to distinguish between different channels, it is made sure that the color of the text in any particular _TextView_ always **syncs** with the color of its parent waveform.

This feature has been developed for and tested with all the channels (i.e., _CH1_, _CH2_, _CH3_, _In-Built MIC_, _PSLab MIC_).

## Screenshots / Recordings  
Here, is a demonstration of the feature:-

https://github.com/fossasia/pslab-android/assets/125425881/6207b9ff-ab9c-437c-83df-2150982b8d68

**Checklist**: 

@CloudyPadmal @marcnause Your views on this ? Any suggestions regarding the layouts or anything else are most welcome.
@marcnause Could you also please test this feature and confirm that everything works fine (there may be very minute errors in the frequencies displayed, these are very normal given the hardware limitations and the number of samples being used to calculate the frequencies and periods) ?

### DIFF
--- a/app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
+++ b/app/src/main/java/io/pslab/activity/OscilloscopeActivity.java
@@ -34,6 +34,9 @@ import androidx.annotation.Nullable;
 import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.widget.Toolbar;
 import androidx.fragment.app.Fragment;
+import androidx.recyclerview.widget.DefaultItemAnimator;
+import androidx.recyclerview.widget.LinearLayoutManager;
+import androidx.recyclerview.widget.RecyclerView;
 
 import com.github.mikephil.charting.charts.LineChart;
 import com.github.mikephil.charting.components.Legend;
@@ -61,6 +64,7 @@ import butterknife.BindView;
 import butterknife.ButterKnife;
 import io.pslab.R;
 import io.pslab.activity.guide.GuideActivity;
+import io.pslab.adapters.OscilloscopeMeasurementsAdapter;
 import io.pslab.communication.AnalyticsClass;
 import io.pslab.communication.ScienceLab;
 import io.pslab.fragment.ChannelParametersFragment;
@@ -76,6 +80,7 @@ import io.pslab.others.CSVLogger;
 import io.pslab.others.CustomSnackBar;
 import io.pslab.others.GPSLogger;
 import io.pslab.others.LocalDataLog;
+import io.pslab.others.OscilloscopeMeasurements;
 import io.pslab.others.Plot2D;
 import io.pslab.others.ScienceLabCommon;
 import io.realm.Realm;
@@ -164,6 +169,8 @@ public class OscilloscopeActivity extends GuideActivity implements View.OnClickL
     TextView xyPlotTextView;
     @BindView(R.id.parent_layout)
     View parentLayout;
+    @BindView(R.id.recyclerView)
+    RecyclerView measurementsList;
     private Fragment channelParametersFragment;
     private Fragment timebaseTriggerFragment;
     private Fragment dataAnalysisFragment;
@@ -184,6 +191,7 @@ public class OscilloscopeActivity extends GuideActivity implements View.OnClickL
     private double maxAmp, maxFreq;
     private boolean isRecording = false;
     private boolean isRunning = true;
+    private boolean isMeasurementsChecked = false;
     private Realm realm;
     public RealmResults<OscilloscopeData> recordedOscilloscopeData;
     private CSVLogger csvLogger;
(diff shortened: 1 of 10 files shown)


CloudyPadmal:
Looks really nice!

marcnause:
Yes, I agree! It looks really good and I think the position is well chosen!

I noticed a few issues. I connected wave 1 to CH1 of the oscilloscope and set a frequency. All other channels of the oscilloscope are not connected. The phone is my Motorola Moto X Play, which is pretty slow by today's standards.

- When I turn on and off the channels, it seems like the screen is updated slower the more channels are active.
- CH2 and CH3 show the same values for frequency and phase as CH1, even though they are not connected (0:39 in the video)
- At some point the correct waveform of CH1 is only displayed if CH2 is active, otherwise it is only a flat line (1:13 in the video)
- The size of the TextView changes depending on the displayed values. (1:13 in the video) This _may_ have a negative effect on the performance of the UI because the system _may_ have layout and redraw more parts of the screen than required. But I am not sure if that is really the case.

I have only had a superficial look at the code and have not done any profiling, so I have no idea where to start looking for the cause yet.

https://github.com/fossasia/pslab-android/assets/11596220/7bc41571-49f2-4131-ae56-7f5c96c7cc0b

AsCress:
@marcnause Thanks very much for the testing and your valuable inputs !! Most of these issues have already been observed by me. Let's just go through these one by one:
> * When I turn on and off the channels, it seems like the screen is updated slower the more channels are active.

This is quite normal for the oscilloscope, and isn't related to just the _Automated Measurements_ feature as the larger the number of channels activated, the more data we have to capture, which also takes additional time. You would observe the lagging irrespective of whether the _Automated Measurements_ is activated or not. It also varies from device to device (e.g.- I tested this on a Motorola One Power device which may not be as old as yours, and the capturing was quite fast). Also, the methods that we use to capture data from the PSLab device are outdated now and I would completely be revamping our commands to make it a literal 1:1 port from the python implementation (see #2490). The lag issues would be solved by that implementation.

> * CH2 and CH3 show the same values for frequency and phase as CH1, even though they are not connected (0:39 in the video)

I had observed this earlier as well, this issue isn't related to the _Automated Measurements_ feature. This seems to be an issue with either the hardware or the firmware. So basically, whenever we configure a particular channel to generate some analog wave of a particular frequency, a signal of the same frequency having very very less amplitude (so small that our algorithm would rule it out as noise) is generated on the other channels as well (if they are **not** generating their own waves). The _Automated Measurements_ feature just takes in whatever input is coming and calculates the characteristics.

> * At some point the correct waveform of CH1 is only displayed if CH2 is active, otherwise it is only a flat line (1:13 in the video)

This again is an issue with the _Oscilloscope_ itself and not the _Automated Measurements_ feature. I doubt if it's an issue in the hardware/firmware but it seems to me that the issue is somewhere in the current command implementation in the Android app, as the wave is generated correctly by the PSLab device but it isn't read properly by the app. Also, this issue isn't just limited to _CH1_ and _CH2_, I was able to reproduce this randomly with all the channels (even the _In-Built MIC_! :)) ). Nevertheless, all this will be solved once I make a PR for #2490 as I mentioned above as well.


> * The size of the TextView changes depending on the displayed values. (1:13 in the video) This _may_ have a negative effect on the performance of the UI because the system _may_ have layout and redraw more parts of the screen than required. But I am not sure if that is really the case.

Really great suggestion !! I've just pushed in a new version in which I have improved the layout as well as configured the _TextViews_ in such a way that they won't resize as much as they were resizing before. I've updated the screen recording in the body of this PR for you to have a look :)).

So, I think that most of these issues are with the _Oscilloscope_ itself and not the _Automated Measurements_ feature. Perhaps, we can open separate issue tickets for these after we merge a PR for #2490 (if they would still be existing).
@marcnause Do inform me if all this makes sense 😂😂.

marcnause:
@AsCress Makes sense. :smiley:

marcnause (approved):

PR #2551, opened 2024-09-17
fix: disable channel range select when in running state

### LINKED ISSUE #2099: Oscilloscope: Range does not change, when option is chosen
The Oscilloscope has the option to change the range, e.g. of the Voltage display. This is currently not working. There is no visual change in the app if the user changes the range.

* Th setting sensitivity doesn't work, it is always +-16V

Screenshot_20200203_133437_io pslab

Dishebh:
Fixes #2099

## Changes 
- disabled channel range select dropdown for CH1 and CH2 when in running state under Oscolloscope

## Screenshots / Recordings  

**Checklist**: 
- [ ] **No hard coding**: I have used resources from `strings.xml`, `dimens.xml` and `colors.xml` without hard coding any value.
- [ ] **No end of file edits**: No modifications done at end of resource files `strings.xml`, `dimens.xml` or `colors.xml`.
- [ ] **Code reformatting**: I have reformatted code and fixed indentation in every file included in this pull request.
- [ ] **No extra space**: My code does not contain any extra lines or extra spaces than the ones that are necessary.

## Summary by Sourcery

Fix the issue where the channel range select dropdown was active during the running state of the oscilloscope by disabling it, and refactor dataEntries to use a more consistent type.

Bug Fixes:
- Disable the channel range select dropdown for CH1 and CH2 when the oscilloscope is in the running state to prevent unintended changes.

Enhancements:
- Refactor dataEntries from ArrayList<ArrayList<Entry>> to List<List<Entry>> for improved type consistency.


Dishebh:
cc @marcnause @AsCress

Dishebh:
hey @marcnause @AsCress , sorry for tagging you again, but if you get a chance to look at this PR, and find any issues, please let me know. Thanks!

marcnause:
@sourcery-ai review

marcnause:
The changes have the described effect. I am just not sure anymore if disabling channel range select is the solution of if it only hides a bug that should be fixed. Please let me investigate and come back to this PR once I have made up my mind.

AsCress:
Thanks for your work on this issue !
My view is that we could possibly have a better way of solving this instead of just disabling the range select when the Oscilloscope is running.
We have a piece of code which sets the axes range whenever a sample is captured from the Oscilloscope. How about not updating the range **everytime**, and just setting it once using the range select spinners ?
Maybe we can work on these lines and solve this issue in a better way @Dishebh.
What do you think of this @marcnause ?

marcnause:
@AsCress I checked the `OscilloscopeActivity` and I guess line 1185 is the piece of code you are referring to, right? I removed lines 1185 and 1186 locally and I am finally able to set the range as expected. Do you think there may be any negative effects of this change?

One additional problem I noticed: If CH1 and CH2 are selected, both are scaled according to the range set for CH1. The range for CH2 is ignored. But maybe we should handle this problem in an additional issue.

marcnause:
I think there will be a problem if the lines are removed and a user switches to FFT mode and back. I think the whole issue is more complicated than I thought when I wrote my first comment in https://github.com/fossasia/pslab-android/issues/2099. I think it would probably make things easier if all variables and methods that are connected to range selection is moved to an extra class. I will try to do some refactoring tonight if I find the time.

AsCress:
@marcnause Hmm, I see. I think there may also be a problem with the _Auto-Scale_. Do inform me if you need any help with this (currently I am working a lot on the _ESP_ and the _Logic Analyzer_, so didn't get time to push any changes to this PR or any other 😢).

marcnause:
I have added a new pull request in which I try to fix the underlying problem: https://github.com/fossasia/pslab-android/pull/2552

AsCress:
@Dishebh Since this is now superseeded by #2552, are you ok with closing this PR ?

Dishebh:
Hey, sure. Let me close this, if Marc's PR already fixes the issue. Thanks, @marcnause, for fixing this, and sorry I couldn't do it 😅

marcnause:
@Dishebh Thank you very much for taking on this problem. Your contribution was the first step towards solving this problem, without it the bug ticket would probably still be open!

PR #2559, opened 2024-10-31
fix: fixed errors in Power Source values

### LINKED ISSUE #2549: Power Supply: Set values have a relatively high offset
**Actual Behaviour**

Setting the values of any of the instruments in the Power Supply screen results in values being set which differ from what the user requested. For PV1 I measured an error of up to 0.15 V (for 1.6V on PV1) with a PSLab v5 board. I also set values for PV1 with the Python library and the maximum error was 0.016 V (see attached Libre Office Calc file [Multimeter-2024-08-26 22-30-59-Pass.ods](https://github.com/user-attachments/files/17005830/Multimeter-2024-08-26.22-30-59-Pass.ods)).

Possibly the problem is worse with the v5 board. I did not have the time for extensive testing with the v6 board, but it seems like the error is smaller (0.03 V for 1.6 V on PV1).

**Expected Behaviour**

Setting PV1, PV2 and PV3  in the app should result in an output which has an error which is not significantly higher than setting the value with the Python library.

**Steps to reproduce it**

1. Connect PV1 to CH1
2. Go to Power Supply
3. Set value for PV1
4. Go to Multimeter
5. Measure voltage at CH1
6. Go to 2. and try for different values

I already played around a little bit, but to no avail. The next thing I'd try would be to compare the bytes being sent by the app and the Python library for certain voltages.

If anybody would like to have a look, that would be great. I don't know if I'll find enough time for this issue soon.

AsCress:
Fixes #2549 

## Changes 
- Errors in setting values using the _Power Source_ were due to duplicate commands being sent to the _PSLab_ for each value change, as the controllers were synced.
- This issue is now resolved: a single command is sent to the _PSLab_ for each user-set value.

## Screenshots / Recordings  
N/A

**Checklist**: 

@marcnause Could you please check and confirm if the values being set now are accurate ?

## Summary by Sourcery

Bug Fixes:
- Resolve issue with duplicate commands being sent to the PSLab for each value change by ensuring a single command is sent for each user-set value.

### DIFF
--- a/app/src/main/java/io/pslab/activity/PowerSourceActivity.java
+++ b/app/src/main/java/io/pslab/activity/PowerSourceActivity.java
@@ -158,6 +158,8 @@ public class PowerSourceActivity extends GuideActivity {
 
     private float voltagePV1 = 0.00f, voltagePV2 = 0.00f, voltagePV3 = 0.00f, currentPCS = 0.00f;
 
+    private Boolean _update = true;
+
     public PowerSourceActivity() {
         super(R.layout.activity_power_source);
     }
@@ -523,12 +525,12 @@ public class PowerSourceActivity extends GuideActivity {
         controller.setMax(controllerLimit);
         controller.setProgress(retrievePowerValues(pin));
         controller.setOnCrollerChangeListener(new OnCrollerChangeListener() {
-            private int progress;
-
             @Override
             public void onProgressChanged(Croller croller, int progress) {
-                setMappedPower(pin, progress);
-                this.progress = progress;
+                displayMappedPower(pin, progress);
+                if (_update) {
+                    setPower(pin);
+                }
                 removeCursor();
             }
 
@@ -537,6 +539,7 @@ public class PowerSourceActivity extends GuideActivity {
 
             @Override
             public void onStopTrackingTouch(Croller croller) {
+                _update = false;
                 setPower(pin);
                 /*
                   V6 hardware has two pairs of paired channels:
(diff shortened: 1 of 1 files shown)


CloudyPadmal (approved):

marcnause:
Hmmm, I still have offsets of approximately 0,2V. Does it work correctly with your devices?

marcnause:
Works much better now!

I noticed that the value in the `TextView` is only updated when the voltage is changed via the buttons or when the user stops touching the controller (`Croller`). Is that intentional?

AsCress:
@marcnause Yes that has been kept intentionally by me. What are your views on this?

marcnause:
I find it a tad confusing if the TextViews are not updated when the controller is changed. I guess the reason for the change was to avoid sending too many commands, right?

AsCress:
@marcnause I think yes you are right. Probably, I should think about another fix for this issue rather than updating the `TextViews` when the user stops touching the `Croller`.
It is indeed confusing. I'll push to this branch soon.

AsCress:
@marcnause This is finally ready for testing now !
Easily, the hardest thing that I've worked upon in the _PSLab_ till now 😂😂.
Syncing in those controllers was damn confusing.
I've tested this with a Multimeter, and I can confirm from my side that the errors now are vanished, and these are the closest values we can get from the _PSLab_. Any error remaining now is because of the hardware limitations of the _PSLab_ device.

marcnause (approved):
This has come a long way! 😉

Issue #2584, opened 2024-12-16
Not all sensors which have been implemented are shown in SensorActivity


marcnause:
**Actual Behaviour**

SensorActivity does not list all sensors. (9 sensors are listed.)

**Expected Behaviour**

SensorActivity should list all available sensors. (11 Actvities for sensors exist.)

The reason why not all sensors are listed is probably that there are duplicate keys in `sensorAddr` in `SensorActivity`.

**Steps to reproduce it**

1. Open app
2. Tap "Sensors"
3. Tap "AUTOSCAN" button

**LogCat for the issue**

N/A

**Screenshots of the issue**

<img src="https://github.com/user-attachments/assets/db2e3199-0d26-4e26-ba40-d72e3070d461" width=400 />
<img src="https://github.com/user-attachments/assets/cc2e35b6-f904-44c4-91de-19d11d8ed667" width=400 />

**Would you like to work on the issue?**

Maybe.

samruddhi-Rahegaonkar:
@marcnause  I would be like to work on this issue.

marcnause:
@samruddhi-Rahegaonkar I have assigned the issue to you. I am not sure what the numbers which are used as keys in in `sensorAddr` in `SensorActivity` mean.

@AsCress You probably added the new sensor implementations during GSoC. Do you remember where the numbers come from?

samruddhi-Rahegaonkar:
@marcnause I am Facing some issues while setting up the project . Why we don't have following code snippet for jitpack in build.gradle.

	dependencyResolutionManagement {
		repositoriesMode.set(RepositoriesMode.FAIL_ON_PROJECT_REPOS)
		repositories {
			mavenCentral()
			maven { url 'https://jitpack.io' }
		}
	}

samruddhi-Rahegaonkar:
@marcnause I am getting following error while building project do you know how to resolve it ?
[Fatal Error] jetifier-core-1.0.0-beta10.pom:2:90: The markup in the document following the root element must be well-formed.

samruddhi-Rahegaonkar:
@marcnause  Solved !👍

samruddhi-Rahegaonkar:
@marcnause Solved Not all sensors which have been implemented are shown in SensorActivity #2584
<img width="1248" alt="Screenshot 2024-12-18 at 1 33 13 PM" src="https://github.com/user-attachments/assets/c83835b7-7bcd-4985-a25f-426590202371" />
All Sensors are showing.

[...]

PR #2589, opened 2024-12-18
fix: Display all implemented sensors in SensorActivity #2584

samruddhi-Rahegaonkar:
Updated the sensor addresses to be unique in the SensorActivity class, resolving conflicts where multiple sensors were assigned the same address (e.g., 0x39 for APDS9960 and TSL2561).
Refactored the sensor initialization to ensure accurate mapping between sensor addresses and their names, allowing all implemented sensors to be correctly shown during the sensor scan.

Screenshot 2024-12-18 at 1 47 43 PM

Checklist:

 No hard coding: I have used resources from strings.xml, dimens.xml, and colors.xml without hard coding any value.

 No end of file edits: No modifications done at the end of resource files strings.xml, dimens.xml, or colors.xml.

 Code reformatting: I have reformatted code and fixed indentation in every file included in this pull request.

 No extra space: My code does not contain any extra lines or extra spaces than the ones that are necessary.

## Summary by Sourcery

Fix sensor address conflicts in SensorActivity by assigning unique addresses to each sensor and refactor initialization for accurate mapping. Update the IndicatorSeekBar dependency version in the build configuration.

Bug Fixes:
- Resolve conflicts in sensor addresses by assigning unique addresses to each sensor in the SensorActivity class.

Enhancements:
- Refactor sensor initialization to ensure accurate mapping between sensor addresses and their names.

Build:
- Update the version of the IndicatorSeekBar dependency in the build configuration.


AsCress:
Hello @samruddhi-Rahegaonkar ! Welcome here!
Thanks for your work on this issue !
Let me provide some context on those duplicate numbers as asked by @marcnause as well in #2584.
These numbers are actually unique addresses for each sensor, which is something which we can't decide upon, but rather is fixed by the sensor manufacturer itself.
This unique address can be found in the datasheet of each respective sensor. The I2C bus identifies anything connected on it using this address.
Since this address is a 7-bit (or 10-bit number), only 128 (or 1024) addresses for I2C sensors are possible, which makes it clear that there would definitely be duplications (there are way more sensors in the world). 
However, this duplication would cause a problem only if two sensors having the same address are connected to the I2C bus (which isn't gonna be the case most of the times with the _PSLab_).
Therefore, we can safely keep the duplications and add those sensors. We can however inform the user, each time the _PSLab_ detects a sensor connection which can possibly have a duplicacy. In such a either one of these sensors would be operable (if more than one with the same address are connected).

AsCress:
As for this pull-request, it doesn't actually solve the mentioned issue of displaying the sensors which arn't listed in the `SensorActivity`.
@samruddhi-Rahegaonkar Possibly you can add those in the list, without changing any of the addresses.
I'll mark the necessary changes in a review to make it easy for you.

app/build.gradle.kts
```diff
@@ -83,7 +83,6 @@ dependencies {
     implementation("com.github.mirrajabi:search-dialog:1.2.4")
     implementation(files("../libs/croller-release.aar"))
     implementation("com.github.BeppiMenozzi:Knob:1.9.0")
-    implementation("com.github.warkiz:IndicatorSeekBar:v2.1.1")
```
AsCress: Some changes which arn't required. You can just rollback `build.gradle.kts.`

app/src/main/java/io/pslab/activity/SensorActivity.java
```diff
@@ -74,17 +74,18 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+        sensorAddr.put(0x48, "ADS1115");  // Unique address for ADS1115
+        sensorAddr.put(0x77, "BMP180");   // Unique address for BMP180
+        sensorAddr.put(0x5A, "MLX90614"); // Unique address for MLX90614
+        sensorAddr.put(0x1E, "HMC5883L"); // Unique address for HMC5883L
+        sensorAddr.put(0x68, "MPU6050");  // Unique address for MPU6050
+        sensorAddr.put(0x40, "SHT21");    // Unique address for SHT21
+        sensorAddr.put(0x39, "TSL2561");  // Unique address for TSL2561
+        sensorAddr.put(0x69, "MPU925x");  // Unique address for MPU925x
+        sensorAddr.put(0x29, "VL53L0X");  // Unique address for VL53L0X
+        sensorAddr.put(0x3A, "CCS811");  // New unique address for CCS811
```
AsCress: As mentioned earlier, changing the address would make the sensor inoperable. You can retain the same address.

app/src/main/java/io/pslab/activity/SensorActivity.java
```diff
@@ -74,17 +74,18 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+        sensorAddr.put(0x3B, "APDS9960"); // New unique address for APDS9960
```
AsCress: Same for this.

build.gradle.kts
```diff
@@ -11,5 +11,5 @@ buildscript {
 }
 
 plugins {
-    id("com.android.application") version "8.7.3" apply false
```
AsCress: Again, some changes that arn't required. You can rollback these.

AsCress (changes requested):

build.gradle.kts
samruddhi-Rahegaonkar: Okay @AsCress Thanks for Explaining me. I will do the needful !

samruddhi-Rahegaonkar: @AsCress Other than the SensorActivity the changes I have made because I was unable to run the application without making those changes. But after testing I will rollback it.

samruddhi-Rahegaonkar: @AsCress to resolve this issue as you have said, I need to make changes in the method "PopulateSensors" to handle the cases where both having same addresses.
should I ?

AsCress: @samruddhi-Rahegaonkar The problem here is with the data structure that we are using to store the sensor addresses.
 A `Map` in Java doesn't allow duplicate keys to be inserted. When we try to do that in lines 86 and 87 of `SensorActivity.java`, it overwrites the previous entries, and hence the previous two sensors (`MLX90614` and `TSL2561`) actually never get inserted in the `Map`. 
The idea here would be to migrate to a different data structure which would allow storage of duplicate entries, and then using that in `PopulateSensors` to display them.

samruddhi-Rahegaonkar: @AsCress Instead of `Map<Integer, String>`, we can use a `Map<Integer, List<String>>` to associate multiple sensor names with the same address.

AsCress: @samruddhi-Rahegaonkar Yep, that's one way. Go ahead and maybe try to implement whatever you find best in this PR.

app/build.gradle.kts
```diff
@@ -83,7 +83,6 @@ dependencies {
     implementation("com.github.mirrajabi:search-dialog:1.2.4")
```
AsCress: Please rollback this whole file as discussed earlier.

app/src/main/java/io/pslab/activity/SensorActivity.java
```diff
@@ -176,31 +179,29 @@ protected Void doInBackground(Void... voids) {
...
+                        dataName.addAll(sensorAddr.get(address));
                     }
                 }
-
-                for (final String s : dataAddress) {
-                    tvData.append(s).append(":").append(sensorAddr.get(Integer.parseInt(s))).append("\n");
-                }
-
-            } else {
-                tvData.append(getResources().getString(R.string.sensor_not_connected));
```
AsCress, line 192: We don't have to get this removed.  `tvData` displayes the addresses of the connected devices. 
Kindly rollback the changes from lines 186 to 192.

app/src/main/java/io/pslab/activity/SensorActivity.java
```diff
@@ -176,31 +179,29 @@ protected Void doInBackground(Void... voids) {
...
+                for (String sensor : sensors) {
+                    if (!dataName.contains(sensor)) {
+                        dataName.add(sensor);
+                    }
+                }
             }
 
             if (scienceLab.isConnected()) {
-                tvSensorScan.setText(tvData);
+                tvSensorScan.setText(getString(R.string.not_connected));
```
AsCress: Not needed. See comment above.

build.gradle.kts
```diff
@@ -11,5 +11,5 @@ buildscript {
 }
```
AsCress: Please rollback as discussed earlier.

AsCress (changes requested):
@samruddhi-Rahegaonkar Thanks again for your work.
There are some things which have to be changed after which this PR would be good to go. 
Marking all these in a review.

build.gradle.kts
samruddhi-Rahegaonkar: @AsCress Sorry for late! please check the commit once. and suggest changes if needed.

AsCress:
@samruddhi-Rahegaonkar Let me clear up some more things here. 
You can never revert changes to files by pushing new commits which don't contain those files. What you are essentially doing doesn't remove those changes from your branch. Changes made through commits to any branch are additive in nature, with every commit you push, your changes get **added** to changes in previous commits.
Thus, if you have a look in the _Files changed_ section in this PR, you would observe that it still shows 3 files, `build.gradle.kts`, `app\build.gradle.kts` and `SensorActivity.java`, something which we don't want to have.
In order to revert changes, you can do any one of these things:
1. Revert the changes manually in the files (delete what you added and add what you deleted) to the files manually and then make a new commit (something which I don't recommend as you would always end up having some changes that arn't required).
2. Perform a Git Soft Reset (`git reset --soft HEAD~n`) which would reset the commit history by removing the last n commits, but would keep your changes. You could soft reset all the commits in this branch, rollback all the files not required, make a new fresh commit and then **force push** this branch.

samruddhi-Rahegaonkar:
@AsCress sorry, I am beginner in OpensSource contribution and Learning by doing. Thanks for your guidance.

samruddhi-Rahegaonkar:
@AsCress Actually when I am trying to revert the changes the gradle sync is not completing.
The project is using an incompatible version (AGP 8.7.3) of the Android Gradle plugin. Latest supported version is AGP 8.5.1
See Android Studio & AGP compatibility options.

AsCress:
@samruddhi-Rahegaonkar The only reason I can think of a failed _gradle_ sync could be an outdated version of Android Studio.
You could update it to the latest version and it would work.
For now, you can just test your build by making your changes in the build files and then just rollback those before committing. We have appropriate workflows in our repository to check your build here, so no worries !

AsCress:
> The project is using an incompatible version (AGP 8.7.3) of the Android Gradle plugin. Latest supported version is AGP 8.5.1
> See Android Studio & AGP compatibility options.

Yep. The issue is Android Studio only. Just update it to the latest version and everything would work fine.

AsCress:
@marcnause This is good to go from my side now ! Maybe you could give it a final review before we merge this.

WhatsApp Image 2025-01-09 at 9 41 58 PM

AsCress (approved):
Nice work @samruddhi-Rahegaonkar ! Thank you very much !!

marcnause (approved):

samruddhi-Rahegaonkar:
@AsCress @marcnause Thank You for Guiding me throughout the issue!
@AsCress Special Thanks to you! for your valuable guidance.

PR #2642, opened 2025-03-10
fix: ESP01 connection

AsCress:
Fixes the connection with ESP01 chip.

## Summary by Sourcery

Implements connection to ESP01 chip via WiFi, including background task to handle socket connection and UI updates based on connection status.


marcnause (approved):
Successfully tested on my device (Fairphone 4, Android 13) and with the PSLab v6 wit the 3.1.0 firmware.

Issue #2678, opened 2025-05-10
Robotic Arm: wrong value may be used to set value if user enters blank value


marcnause:
**Actual Behaviour**
A cached value is use if the user enters a value for one servo and enters a blank value for another value.

**Expected Behaviour**
The value for a servo should always be the value of the same servo before the user enters a blank value, not a value which was entered previously.

**Steps to reproduce it**
0. Open app
1. Open screen "Robotic Arm"
2. Tap text field of servo 1 and enter a value 
3. Tap text field of servo 2 and remove the value
4. Tap enter (green button)

Value of servo 2 will be set to same value as servo 2.

**Screenshots of the issue**

https://github.com/user-attachments/assets/b8552104-913c-4e35-85e0-a71376a129b9

**Would you like to work on the issue?**

Yes, but feel free to grab this issue as long as I have not assigned it to myself.

Anxhul10:
I am able to reproduce this when servo1 value is suppose set to 100 then removing the sevo2 value and then saving gives 100. 

now suppose servo1 value is set to 200 then again removing servo 2 and then saving it it gives 100.

PR #2689, opened 2025-05-15
fix: rounding error in robotic arm

### LINKED ISSUE #2677: Robotic Arm: rounding error when entering values with keyboard
**Actual Behaviour**
When a value is set for a servo and a different value is set via keyboard, it can happen that the displayed value is not equal to the entered value.

**Expected Behaviour**
The displayed value is always the same value as the one entered by the user.

**Steps to reproduce it**
0. Open app
1. Tap on "Robotic Arm"
2. Change the value of Servo 1 by dragging the blue dot along the circle to 100
3. Tap in the middle of the circle (keyboard should open)
4. Tap a second time if no cursor is visible
5. Enter 38 as the new value
6. Tap green Enter key

--> Value is magically changed to 36.

**Screenshots of the issue**

https://github.com/user-attachments/assets/5d68f122-38fc-4435-b872-60363ec23f0e

Yes, but feel free to grab this issue as long as I have not assigned it to myself.

rahul31124:
Fixes #2677 #2678

## Changes 
The error was due to the Triggertrap SeekArc library’s setProgress method accepting only integer values, and the getProgress method returning only integers. This caused incorrect percentage calculations for corresponding degree values.

I have now added a customized SeekArc class by copying the existing class from the Triggertrap SeekArc library, as the library is currently inactive. The customized class includes setProgress and getProgress methods that can accept and return floating-point values, respectively.

## Screenshots / Recordings  
[Robotic_Arm_Rounding.webm](https://github.com/user-attachments/assets/c8f59541-8f94-4561-86eb-b6a250b5b004)

**Checklist**: 

## Summary by Sourcery

Introduce a custom circular seek bar with floating-point progress to fix rounding errors and migrate existing references to it across the Robotic Arm UI.

New Features:
- Add CSeekArc component supporting float precision for progress values.

Bug Fixes:
- Fix incorrect degree calculation by rounding float progress instead of using integer-only API.

Enhancements:
- Replace all original SeekArc usages with CSeekArc in activity code and XML layout.
- Update progress change listeners and degree input handling to use float progress and round results.


AsCress:
@rahul31124 Could you please fix the errors reported by the Codacy SCA before we further in reviewing this ?

rahul31124:
Hello @AsCress, the errors reported by Codacy SCA have now been fixed

app/src/main/java/io/pslab/others/CSeekArc.java
```diff
@@ -0,0 +1,551 @@
+/*******************************************************************************
```
AsCress: Why copy the whole code here ? 
The ideal thing to do would be just to create a new instance of `SeekArc`, and override/implement the methods that you want to change.
This is never how we make amendments to libraries.

AsCress (changes requested):
@rahul31124 Nice work there ! One fundamental thing to correct. Rest everything looks nice !

app/src/main/java/io/pslab/others/CSeekArc.java
rahul31124: @AsCress ,
Thanks for the feedback! I understand that copying the entire code isn’t ideal. However, since SeekArc is from an external library and some important methods/variables like mProgress, which I changed from int to float are private. As far as I know, we can’t override private fields or methods, which is why I ended up modifying the entire class

AsCress: Then, the ideal thing would be to create new variables and implement only those methods which are required / affected by those variables. That's precisely why I mentioned that you can also **implement** the ones which arn't overridable.

rahul31124: @AsCress,
Since many of the variables and methods in SeekArc are private or tightly coupled internally, introducing new variables would still force us to rewrite or duplicate most of the class. For example, even within a public method, a private method might be called, which again uses private variables so we end up having to redefine those as well and there very few public method almost every is private. That’s why I think maintaining a clean copy with our necessary changes might be a more practical and manageable approach

CloudyPadmal:
What prevents the use of the SeekArc package directly, as mentioned in [this link](https://github.com/neild001/SeekArc?tab=readme-ov-file#to-use-it-in-your-code), instead of having to copy and paste the entire class?

CloudyPadmal:
Can’t we work around the integer issue by representing angles as multiples of 10? 

For example, instead of using 360 degrees, we can set the seek bar’s maximum value to 3600 and interpret values accordingly, treating 3600 from the seek bar as 360 degrees in code.


This might seem like a bit of an overkill, but would something like this work as a simple fix?

```
seekArc.setMax(3600);
float angle = 123.4f;
seekArc.setProgress((int)(angle * 10));
float angle = seekArc.getProgress() / 10.0f;
```

rahul31124:
Hello @CloudyPadmal
, I believe the setMax method was added after version 1.1 of the library. Since no newer version has been released since 1.1, we're unable to access this method through the current version, even though it exists in their github codebase
Screenshot 2025-05-16 100345

CloudyPadmal:
I see. The issue seems that the progress level is set to 100 as a private variable with no handler to change it.


There are probably two options,
- Find a clever way to change this `mMax` value to 360 instead of 100.
- Use a different library since this is depreciated.

rahul31124:
@CloudyPadmal, Should I fork the repo, modify the existing code, and create my own version of the library? Would that be a better approach?

CloudyPadmal:
I just found a forked version https://github.com/simonebortolin/MaterialSeekArc/tree/master

Will that work?

rahul31124:
@CloudyPadmal, I’ve tested this version of the MaterialSeekArc, but unfortunately, I believe it may still not be suitable for use. The project seems inactive, and there are some key issues I've encountered during testing:

The drawing of the SeekArc is not working correctly  the arc either doesn’t render and touch interaction is also unresponsive

AsCress:
> @CloudyPadmal, Should I fork the repo, modify the existing code, and create my own version of the library? Would that be a better approach?

@rahul31124 I would suggest to maybe have a look at other libraries which can solve this purpose. However, if you don't find one then this could be a way forward. If you want to use a library in your project which is deprecated and needs fixing / needs some new features which you want to work on and the library owners don't accept new contributions, there are broadly only two approaches possible:

1. Clone the library repository, make your changes, and include the whole code in our repository. _Gradle_ can then be configured to build the whole library along with our code every time you build the app, and use it directly.
2. Fork the repository, make your changes, and configure _Gradle_ to fetch the library from your fork.

rahul31124:
@AsCress I have forked the repository, applied the required changes to the method, and configured the project to use my modified version of the library

AsCress:
@rahul31124 Nice ! Could you please resolve the merge conflicts here ?

AsCress (approved):
LGTM @rahul31124 !

PR #2784, opened 2025-07-12
feat: added l10n for German

AsCress:
Adds localizations for Deutsch to the app.
⚠️ **ALL THESE ARE AUTOMATICALLY GENERATED USING A MACHINE TRANSLATOR**, as I don't know Deutsch (a bit maybe, but not that much 😅).

@marcnause You are the only one who can review these changes here I guess 😂. You can have a look at the `app_de.arb` file.

## Summary by Sourcery

Add German translations and enable German locale support in the app

New Features:
- Include app_localizations_de.dart with German translations
- Register German locale in supportedLocales and localization delegate


lib/l10n/app_de.arb
```diff
@@ -0,0 +1,309 @@
...
+  "yAxisRange3V": "+/-3V",
+  "yAxisRange2V": "+/-2V",
+  "yAxisRange1_5V": "+/-1,5V",
+  "yAxisRange1V": "+/-1V",
+  "yAxisRange500mV": "+/-500mV",
+  "yAxisRange160V": "+/-160V",
+  "channel1": "CH1",
+  "channel2": "CH2",
+  "channel3": "CH3",
+  "mic": "MIKROFON",
```
AsCress: @marcnause In all places wherever there is MIC, it is getting translated to MIKROFON. Do we have a short form for MIKROFON in German, as introducing such a long name for channels, distorts UI in some places.

AsCress: Discussed in meet, defaulting to MIC.

marcnause: Seems like I was too much in "English mode" during the meet to remember that "Mikro" is used as a short form of "Mikrofon" in German. Not sure if is short enough to fit the UI. If not, we should indeed default to MIC.

AsCress:
@marcnause Apart from the MIC translation, are we good to go ahead with these German translations ?

marcnause:
> @marcnause Apart from the MIC translation, are we good to go ahead with these German translations ?

I will make a few changes. The translation is fine in general, but some expressions sound a bit stiff.

AsCress:
@marcnause Sure ! Thank you very much !

marcnause:
I have committed all my changes. I think a few of the answers in the FAQ may be outdated once the flutter app gets released, but I think that is out of scope for this PR.

I will add @mariobehling as a reviewer since he is also a native German speaker.

marcnause:
Just one more question: Is the ARB file the only file I need to change and the Dart files are generated from it or do I also need to change the Dart files?

AsCress:
@marcnause After changing the ARB file, you would have to run `flutter pub get` or just maybe run the app. That would update the dart files automatically.

marcnause:
OK, now now I have really committed all required files (I hope). 

Here is a screenshot of the app on my "problematic" device. 
Looking at it I just noticed that the title of the screen is not translated. Can you fix that in the scope of this PR?

Screenshot_20250718-194108

marcnause (approved):

mariobehling (commented):
There are too many mistakes in this file. It does not make sense to contribute a file if you dont have basic knowledge in the language. A German speaker or someone with some German knowledge should make a PR for that.

PR #2790, opened 2025-07-17
fix: fix data acquisition from TSL2561 luminosity sensor

### LINKED ISSUE #2783: Sensor data of TSL2561 is not captured
Actual: Data of TSL2561 is not displayed in the corresponding screen.

Expected: Data of TSL2561 is displayed correctly.

marcnause:
Fixes #2783

## Changes 
- Try to find sensor by checking all possible sensor IDs (0x29, 0x39, 0x49)
- Fix data acquisition
- Display sensor IDs in hex format in Sensors screen

I think there are still a few things to fix in the screen, but data acquisition works, which was the scope of the issue.

## Screenshots / Recordings 
https://github.com/user-attachments/assets/2c104bca-7251-427d-94c2-fbf6bb2cda40

**Checklist**: 

## Summary by Sourcery

Add support for detecting TSL2561 sensors at multiple I2C addresses and correct raw luminosity reads, while updating the sensor screen to display addresses in hex

New Features:
- Detect TSL2561 devices by probing all potential I2C addresses (0x29, 0x39, 0x49)
- Display sensor addresses in hexadecimal format in the Sensor screen

Bug Fixes:
- Fix raw data acquisition for full spectrum, infrared, and visible channels using correct read commands and bit operations

Enhancements:
- Refactor TSL2561 class to centralize constants, use byte types, and streamline enable/read logic
- Update SensorActivity to use integer lists for addresses and rebuild scan output accordingly

### DIFF
--- a/app/src/main/java/io/pslab/communication/sensors/TSL2561.java
+++ b/app/src/main/java/io/pslab/communication/sensors/TSL2561.java
@@ -52,13 +52,17 @@ public class TSL2561 {
         // set timing 101ms & 16x gain
         if (scienceLab.isConnected()) {
 
-            /* Here we check if any of the possible addresses has an ID in the expected format.
-             * If no sensor is available with teh given ID, the value 0xFFFFFFFF will be returned.
-             * If a different value is returned, we can check for teh expected value (see data sheet
+            /* Here we probe if any of the expected I2C addresses has an ID in the expected format.
+             * If no sensor is available with the given ID, the value 0xFFFFFFFF will be returned.
+             * If a different value is returned, we can check for the expected value (see data sheet
              * for details).
              */
             for (byte addr : ADDRESSES) {
                 address = addr;
+                /* We disable the sensor before probing since it will not return the expected value
+                 * if it is still active due to a previous data capture.
+                 */
+                disable();
                 Log.d(TAG, "Checking address 0x" + Integer.toHexString(address));
                 int id = i2c.readByte(address, REGISTER_ID);
                 if (id != 0xffffffff && (id & 0x0A) == 0x0A) {
--- a/app/src/main/java/io/pslab/communication/sensors/TSL2561.java
+++ b/app/src/main/java/io/pslab/communication/sensors/TSL2561.java
@@ -2,123 +2,148 @@ package io.pslab.communication.sensors;
 
 import android.util.Log;
 
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/pslab/communication/sensors/TSL2561.java
```diff
@@ -2,123 +2,130 @@
...
         this.i2c = i2c;
         // set timing 101ms & 16x gain
         if (scienceLab.isConnected()) {
+
+            /* Here we check if any of the possible addresses has an ID in the expected format.
+             * If no sensor is available with teh given ID, the value 0xFFFFFFFF will be returned.
+             * If a different value is returned, we can check for teh expected value (see data sheet
+             * for details).
+             */
+            for (byte addr : ADDRESSES) {
```
AsCress, line 60: @marcnause Isn't this handled before in the sensor auto scan ? If the auto scan detects the `TSL2561`, it would show up in the list, which would make this screen accessible to the user. By the time, the user reaches this screen, would it even be possible that the TSL2561 isn't found ?
Is this for handling the conflicting addresses supported by TSL2561 (`0x29` and `0x39`), cause as far as I remember in my discussion with @bessman, there's no way of finding out which sensor is present on a conflicting address rather than trying to read data from it. So, if we're going ahead on that path, do you think that we should implement this check in the higher layers itself (in auto scan), to rule out the showing up of the sensors not present itself ? What are your views on this ?

AsCress (commented):
@marcnause Thank you very much for this !

app/src/main/java/io/pslab/communication/sensors/TSL2561.java
marcnause: I don't know if all or at least most sensors have a way of querying an ID or any other distinct value which lets us decide which sensor is present for a certain address. If this was the case, then I think it would make sense to add some intelligence to the auto scan.

I decided to put the detection into the code for the TSL2561 since currently we don't forward the address of a sensor from the Sensor**s** screen to the sensor specific screens. I also did not want to put too much effort into the legacy version, I only wanted to provide a template for the Flutter version which is at least doing the bare minimum of what is required. I only stumbled over the different possible addresses since the sensor I bought does not use the default address.

AsCress: @marcnause Makes sense. We'll take a call on where to implement these checks in the flutter app. You're right, for the Java app this is fine as it just as to act as a template for the flutter port.

AsCress (approved):

PR #2846, opened 2025-08-15
fix: build fails due to deprecated members

marcnause:
I replaced the deprecated code to fix the build.

## Summary by Sourcery

Replace deprecated radio and switch widget properties to restore build compatibility

Bug Fixes:
- Fix build failures caused by deprecated Radio and Switch members

Enhancements:
- Refactor Radio and RadioListTile usage to wrap them in RadioGroup with groupValue and onChanged handlers
- Update Switch widgets to use activeThumbColor instead of deprecated activeColor


marcnause:
I will create a second PR from a clean branch.

PR #2914, opened 2025-09-09
fix: recording stop button

### LINKED ISSUE #2910: Stop button missing on all recording options
There is no stop button that appears when the user presses "Record". Expected while the user is recording a stop button should appear.

Image

rahul31124:
Fixes #2910

Added a new stop recording PNG icon and updated the condition to switch between the record and stop icons correctly

## Recording
[stop_button.webm](https://github.com/user-attachments/assets/0cb9caea-e102-4727-bb6a-e6933ec185e2)

## Summary by Sourcery

Update the recording button to switch between the record and stop icons based on the recording state

Bug Fixes:
- Show the stop-recording icon instead of the record icon when recording is active

Enhancements:
- Add a new asset path for the stop-recording PNG icon


AsCress (approved):

AsCress:
@rahul31124 Let's do this for all the instruments in the scope of this PR only.

rahul31124:
@AsCress Yeah, I’ve updated the common scaffold, so the change will apply to all instruments

AsCress:
@rahul31124 Ahh ok. Got you ! Thank you very much !

PR #2923, opened 2025-09-15
fix: Connect Device screen

### LINKED ISSUE #2911: Connect device button not working
Clicking the connect device button does not open the option to connect the hardware.

Expected:
* Introductory text with option to connect device should be visible
* Device option, e.g. V5, V6 should be visible with basic guide how to identify the hardware
* Option to connect device via USB or Bluetooth should be available

Image

AsCress:
Fixes #2911
Adds initialization of connection to the board upon clicking the 'Connect Device' button.
Copy changes in the _Connect Device_ screen.

## Screenshots / Recordings
[untitled.webm](https://github.com/user-attachments/assets/74a30105-6982-46b7-ace5-0da3d52c8f4a)

## Summary by Sourcery

Initialize device connection on 'Connect Device' action, refine navigation and UI alignment on the Connect Device screen, and bulk-expand localization with updated copy and new keys for control labels, barometer experiments, and playback controls.

New Features:
- Bulk-add new localization keys for control panel labels (Data, Configure, Set Gain, Set Channel, Set Rate, mV), barometer experiment instructions, and playback controls across all supported locales.

Bug Fixes:
- Initialize the board connection when the Connect Device button is pressed.

Enhancements:
- Update navigation logic to reset or navigate correctly to the Connect Device screen.
- Center-align step instructions and update their copy to specify cable types and correct OTG wording.


mariobehling (changes requested):
Update the description on the connection screen, e.g. use USB instead of Micro-USB and add the info that it is possible to connect via Bluetooth if the PSLab has the Bluetooth option.

mariobehling:
Also note that the top bar has a black font color instead of white. I think this should probably be solved in a separate PR.

AsCress:
@mariobehling I'll check this for all instruments and solve separately.

AsCress:
@mariobehling Made the changes as follows:
Screenshot_1758117286

lib/l10n/app_en.arb
```diff
@@ -86,9 +86,10 @@
     "deviceConnected": "Device Connected Successfully",
     "noDeviceFound": "No USB Device Found",
     "stepsToConnectTitle": "Steps to connect the PSLab Device",
-    "step1ConnectMicroUsb": "1. Connect a micro USB(Mini B) to PSLab",
-    "step2ConnectOtg": "2. Connect the other end of the micro USB cable to a OTG",
-    "step3ConnectPhone": "3. Connect the OTG to the phone",
+    "step1ConnectMicroUsb": "1. Connect a USB cable to PSLab.",
+    "step2ConnectOtg": "2. Connect the other end of the USB cable to an OTG.",
```
marcnause, line 90: Connecting the v6 board to a modern phone or tablet with USB-C works without OTG adapter for me. Is it teh same for you? If yes, maybe we should rephrase the instructions.

AsCress: Works with some USB-C cables but not with others in my case. 
I guess OTG is always a safer option.

marcnause: Then let's keep it the way it is.

marcnause (approved):

mariobehling (approved):

PR #2924, opened 2025-09-16
fix: prevent segfault on Linux if device isn't opened

### LINKED ISSUE #2926: Linux app segfaults in the absence of USB permissions
If the Linux app isn't run using root privileges, it doesn't have the permissions to access the raw USB nodes and segfaults as follows:

Image

## Expected behavior:
The app should gracefully show an error but in no case should crash or segfault.

AsCress:
Fixes #2926

## Summary by Sourcery

Prevent segmentation fault on Linux by verifying that the USB device opens successfully before configuring it

Bug Fixes:
- Check the result of mDevice.open() in DesktopUSBCommunicationHandler and return early if it fails to prevent segfault on Linux
- Log an error when opening the USB device fails

### DIFF
--- a/lib/communication/handler/desktop_comms_handler.dart
+++ b/lib/communication/handler/desktop_comms_handler.dart
@@ -61,7 +61,10 @@ class DesktopUSBCommunicationHandler implements CommunicationHandler {
       throw Exception("Device not connected");
     }
     UsbSerialDevice.setAutoDetachKernelDriver(true);
-    await mDevice?.open();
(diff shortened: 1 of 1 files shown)


mariobehling:
Is there an issue about this, which one? Please stick to best practices https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia/

AsCress:
@mariobehling #2926

marcnause (approved):

PR #2927, opened 2025-09-17
feat: add geolocation logging

### LINKED ISSUE #2909: Enhance Luxmeter design and add map feature
The Luxmeter still has a few issues. Please resolve the following:
* [ ] After pressing recording, there is no stop button. Compare to previous version a stop button appeared after a user hit recording, hence indicating that recording is in progress.
* [ ] The default update of the Lux meter data was faster. Implement the same update time as a default as before.
* [ ] Logged Data displays an image called .csv, but there is no option to download or save the image and CSV elsewhere. Only the sharing feature enables to transfer this.
* [ ] The Lux Meter Configurations shows a Black Arrow instead of a white one.
* [ ] When stopping the recording the file name must be filled in by the user. Expected: There should be a default file name same as in the previous app version that e.g. includes date and time etc.

Image
Image
Image

AsCress:
Resolves #2909 
Adds Geolocation logging and map feature for display of logged location for all the instruments for which data logging is enabled.
Uses [OpenStreetMaps](https://osmfoundation.org/) for displaying the location.

## Screenshots / Recordings
Uploading a Google drive link as video is too big to be attached here:
https://drive.google.com/file/d/1_cQioPPqLJRzA8-YBxL19MhVIErzg2Bt/view?usp=sharing

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -1,5 +1,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
+    <uses-permission android:name="android.permission.ACCESS_COURSE_LOCATION"/>
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_LOCATION" />
     <uses-permission android:name="android.permission.RECORD_AUDIO" />
     <uses-feature android:name="android.hardware.usb.host" />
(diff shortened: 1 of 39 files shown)


AsCress:
Some screenshots for quick reference (using the luxmeter as an example): 
Screenshot_1758127334
Screenshot_1758171326
Screenshot_1758171340
Screenshot_1758127343

AsCress:
_Logged Data_ screen showing data logs for all instruments:

[untitled.webm](https://github.com/user-attachments/assets/b51a23a3-3f22-4c5e-83e9-cbd719674917)

marcnause (approved):

PR #2964, opened 2025-10-12
chore: replace USB Serial library with version from pub.dev

### LINKED ISSUE #2962: usb_serial is pulled from git instead of pub.dev
Replace usb_serial with version from pub.dev if possible.

⚠️ Test not only if dependency can be resolved, but also test if app is still working as expected. ⚠️

marcnause:
Fixes: #2962

Requires testing with phone and PSLab device.

## Summary by Sourcery

Build:
- Switch usb_serial dependency from Git master to version ^0.5.0 on pub.dev

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1029,12 +1029,11 @@ packages:
   usb_serial:
     dependency: "direct main"
     description:
-      path: "."
-      ref: master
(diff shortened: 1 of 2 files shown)


marcnause:
I have tested the APK with my phone and the PSLab v6 board. As far as I can tell, communication worked without problems.

CloudyPadmal (approved):

PR #3020, opened 2026-01-02
fix: restrict Power Source inputs to number pad only

### LINKED ISSUE #3019: UX: Power Source input fields show full keyboard instead of number pad
**Describe the bug**
The text fields in the Power Source instrument (PV1, PV2, PV3, PCS) currently open the full QWERTY keyboard. This allows invalid characters to be typed and is inconvenient for entering numeric voltage/current values.

**To Reproduce**
1. Open the Power Source instrument.
2. Tap on any voltage input field (e.g., PV1).
3. Notice that the full text keyboard appears (with letters).

**Expected behavior**
The keyboard should be restricted to the Number Pad (digits and decimal point only) to prevent invalid input and improve usability.

Inuth0603:
**Related Issue**
Fixes #3019

**Describe the solution**
Updated `lib/view/power_source_screen.dart` to use `TextInputType.numberWithOptions(decimal: true)` for:
- PV1
- PV2
- PV3
- PCS

This ensures the user sees a numeric keypad instead of the full QWERTY keyboard, preventing invalid text input and improving UX.

**Screenshots**
Screenshot_20260102-161545

## Summary by Sourcery

Bug Fixes:
- Ensure PV1, PV2, PV3, and PCS input fields present a numeric keypad with decimal support instead of a full keyboard to avoid invalid text input.

### DIFF
--- a/lib/view/power_source_screen.dart
+++ b/lib/view/power_source_screen.dart
@@ -269,13 +269,12 @@ class _PowerSourceScreenState extends State<PowerSourceScreen> {
                             ),
                             textAlign: TextAlign.center,
                             onSubmitted: (value) async {
-                              String powerValue =
-                                  value.replaceAll("V", "").trim();
                               double parsedValue =
-                                  double.tryParse(powerValue) ?? 0.0;
+                                  double.tryParse(value) ?? 0.0;
                               await provider.setPV1(parsedValue);
                             },
                             decoration: InputDecoration(
+                              suffixText: ' V',
                               enabledBorder: OutlineInputBorder(
                                 borderSide: BorderSide(
                                   color: powerSourceBorderLightRed,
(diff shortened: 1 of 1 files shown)


marcnause:
Hello @Inuth0603, thank you for creating the issue and working on the fix. I have tested it and it works fine on my device, but now the units (V, mA) are missing. Is there any way to keep the units with the number pad as input method? If not, can you please add the units to the layout outside of the text box?

Inuth0603:
> Hello @Inuth0603, thank you for creating the issue and working on the fix. I have tested it and it works fine on my device, but now the units (V, mA) are missing. Is there any way to keep the units with the number pad as input method? If not, can you please add the units to the layout outside of the text box?

Thanks for testing! Yes, I removed the units from the controller text because they conflict with the numeric input type (the keypad doesn't allow typing letters back in if they get deleted).

I will push a fix immediately to add the units back using the suffixText property in InputDecoration. This will keep them visible in the layout to the right of the number, but keep the input field strictly numeric.

marcnause:
> I will push a fix immediately to add the units back using the suffixText property in InputDecoration. This will keep them visible in the layout to the right of the number, but keep the input field strictly numeric.

👍 That sounds like a great solution!

Inuth0603:
Fixed! I have updated the code to use suffixText for the units.

I tested it on a real device:

The numeric keypad now opens correctly.

The units (V, mA) are visible on the right side.

The input logic works perfectly without crashes.

Ready for review!

marcnause (approved):

marcnause:
Thank you for the latest changes! I tested it on my Android device and I also built and tested a Linux version to test what happens if I input invalid values with the hardware keyboard of my computer. It falls back to a default value (0.00) which I guess should be expected and desired behavior.

PR #3047, opened 2026-02-02
fix: Enable resizeToAvoidBottomInset in CommonScaffold

Inuth0603:
### Problem
When users tap on the "Number of Readings" TextField in the `SensorControlsWidget` (located at the bottom of sensor screens), the on-screen keyboard completely covers the input field, making it impossible to see what's being typed.

This affects **all sensor screens** that use `CommonScaffold`:
* APDS9960, Accelerometer, Gyroscope, Barometer, Luxmeter, Thermometer, and more.

### Root Cause
`resizeToAvoidBottomInset` was explicitly set to `false` in `CommonScaffold`, which prevents Flutter from automatically resizing the screen content when the keyboard appears.

## Solution
Removed the explicit resizeToAvoidBottomInset: false line, allowing the scaffold to use Flutter's default behavior (true), which automatically resizes content when the keyboard opens.

## Testing
* Verified on sensor screens with `SensorControlsWidget`.
* Confirmed no overflow issues occur on other screens using `CommonScaffold` (all use scrollable containers like `SingleChildScrollView`, `ListView`, or `GridView`).

## Summary by Sourcery

Bug Fixes:
- Fix bottom-aligned inputs being covered by the keyboard on sensor screens using CommonScaffold.

### DIFF
--- a/lib/view/widgets/common_scaffold_widget.dart
+++ b/lib/view/widgets/common_scaffold_widget.dart
@@ -45,7 +45,7 @@ class _CommonScaffoldState extends State<CommonScaffold> {
   Widget build(BuildContext context) {
     return Scaffold(
       backgroundColor: scaffoldBackgroundColor,
-      resizeToAvoidBottomInset: false,
+      resizeToAvoidBottomInset: true,
       appBar: AppBar(
         systemOverlayStyle: SystemUiOverlayStyle(
           statusBarColor: appBarColor,


Inuth0603:
Hi @marcnause, thank you for requesting the review! I've addressed the Copilot feedback and set resizeToAvoidBottomInset: true explicitly for clarity. The branch has also been updated to sync with upstream. Please let me know if anything else is needed!

rahul31124 (approved):
LGTM!

marcnause (approved):

Issue #3069, opened 2026-02-14
Sound Meter stops updating after returning from configuration screen


bhavjsh:
## Problem
The **Sound Meter** instrument stops updating its real-time data once the user navigates to the **Sound Meter Configuration** screen and returns using the back button.

## Steps to Reproduce
1. Open the **Sound Meter** instrument.
2. Navigate to the **Configuration** screen.
3. Press the **Back Button** (system or AppBar).
4. Observe that the Sound Meter UI becomes static and stops responding to audio input.

## Behavior Comparison

| Expected Behavior | Observed Behavior |
| :--- | :--- |
| The Sound Meter should continue functioning normally after returning. | The instrument becomes unresponsive and stops updating data. |

https://github.com/user-attachments/assets/e295a28b-e3dd-4b83-8230-a53bf85732e0

PR #3071, opened 2026-02-14
fix(soundmeter): prevent rebuild on back navigation

bhavjsh:
**Fixes:** #3069

## Summary
This PR resolves a critical bug where the **Sound Meter** instrument stopped updating real-time data after a user navigated to the Configuration screen and returned.

## Root Cause
The navigation logic within `soundmeter_config_screen.dart` was using complex conditional route handling (such as `pushNamedAndRemoveUntil` or `popUntil`) instead of a standard back-stack pop. 

**This led to:**
-**Unintended Rebuilds:** The Sound Meter screen was being re-initialized rather than resumed.
-**Stream Interruption:** The active audio stream/listener was lost or reset during the route manipulation.
-**UI Freezing:** The instrument state became unresponsive upon returning.

## Changes Made
* **Refactored Navigation:** Replaced custom/heavy route logic with a clean `Navigator.maybePop(context);`.
* **State Preservation:** Ensured the existing Sound Meter screen instance remains in the widget tree without being disposed of or rebuilt.
* **Stream Stability:** Prevented the audio listener from being interrupted by unnecessary route changes.

## Result
* [x] Sound Meter continues updating normally after returning from config.
* [x] Audio streams remain active and uninterrupted.
* [x] Instrument state is fully preserved.

[FIXEDsm.webm](https://github.com/user-attachments/assets/c83b6e7c-33be-4f35-a545-97fda4037b59)

> **Technical Note:** Using `maybePop` ensures that we don't accidentally pop the last remaining route in the stack, providing a safer navigation experience than a forced `pop`.

### DIFF
--- a/lib/view/soundmeter_config_screen.dart
+++ b/lib/view/soundmeter_config_screen.dart
@@ -9,21 +9,15 @@ import '../theme/colors.dart';
 
 class SoundMeterConfigScreen extends StatefulWidget {
   const SoundMeterConfigScreen({super.key});
+
   @override
-  State<SoundMeterConfigScreen> createState() => _SoundMeterConfigScreenState();
+  State<SoundMeterConfigScreen> createState() =>
+      _SoundMeterConfigScreenState();
 }
 
-class _SoundMeterConfigScreenState extends State<SoundMeterConfigScreen> {
+class _SoundMeterConfigScreenState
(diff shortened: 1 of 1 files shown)


rahul31124 (approved):
LGTM!

marcnause (approved):

PR #3100, opened 2026-02-28
pslab app

Neerajgundam-12:
Now pslab app supports multiple platforms I have made the changes now it does not limit it's usability it enables support for Android, iOS, desktop (Windows, macOS, Linux), and web.


marcnause:
Could you please create an issue at https://github.com/fossasia/pslab-app/issues to add a little context to this PR?

marcnause:
Unclear which exact problem is targeted here due to missing issue. Also does not compile.

PR #3134, opened 2026-03-22
Fix pulse count parsing using correct packet handler (getInt instead …

### LINKED ISSUE #3102: Count Pulse Mode Always Returns 0
**Describe the issue**
The pulse count toggle in the Multimeter (Count Pulse) is not showing any value and always displays 0, regardless of the input frequency. For example, when generating a 1200 Hz square wave on SQ1 and connecting it properly to the measurement pin (LA1), the pulse count still remains 0.

I verified the following:

The UI works correctly and the toggle switches properly between Pulse Frequency and Pulse Count modes.

Frequency measurement works correctly.

However, when switching to pulse counting mode, the value always remains 0.

I also tested this behavior in:

The PSLab Python library and in the previous native app.

In both cases, pulse counting consistently returns 0, while frequency measurement works as expected.

Since this behavior is consistent across the platforms, the issue might be related to the firmware.
Maybe, depending on guidance.

Ayushkumar-cybe:
Fixes #3102

## Problem
Pulse count mode in the multimeter always returned `0`, even when valid input signals were present.

## Root Cause
The response from the `fetchCount` command was being parsed using `getVoltageSummation()`, which is designed for ADC/voltage-related data and expects a 3-byte response format.

However, pulse count is a raw numeric value and should be parsed using a standard integer reader.

This mismatch caused incorrect decoding of the device response, leading to incorrect (zero) values.

## Fix
Replaced the incorrect parsing method:
- it is a wrong method is used here `getVoltageSummation()`
- replace getVoltageSummation() to this `getInt()`

This aligns the pulse count response handling with the expected raw integer format returned by the device.

## Changes
- Updated `readPulseCount()` in `science_lab.dart` to use `getInt()` instead of `getVoltageSummation()`

## Screenshots / Recordings
N/A (hardware-dependent feature)

### DIFF
--- a/lib/communication/science_lab.dart
+++ b/lib/communication/science_lab.dart
@@ -858,8 +858,9 @@ class ScienceLab {
     try {
       mPacketHandler.sendByte(mCommandsProto.common);
       mPacketHandler.sendByte(mCommandsProto.fetchCount);
-      int count = await mPacketHandler.getVoltageSummation();
-      return 10 * count;
+
+      int count = await mPacketHandler.getInt();
+      return count;
     } catch (e) {
       logger.e("Error in readPulseCount: $e");
     }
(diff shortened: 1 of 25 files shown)


marcnause:
The issues was fixed in https://github.com/fossasia/pslab-app/pull/3133.

PR #3141, opened 2026-03-27
chore: simplify Dependabot config file

marcnause:
Does not fix # 3124 (see Rahul's comment below), only some minor cleanup.

## Changes
- removed `enable-beta-ecosystems`, it is currently not used according to https://docs.github.com/de/code-security/reference/supply-chain-security/dependabot-options-reference#enable-beta-ecosystems-
- simplify directories according to https://deku.posstree.com/en/flutter/github/dependabot/

## Screenshots / Recordings
N/A

**Checklist**: 

## Summary by Sourcery

Simplify Dependabot configuration for dependency updates.

Enhancements:
- Remove the unused enable-beta-ecosystems option from the Dependabot configuration.
- Consolidate and normalize monitored directories for Gradle and GitHub Actions dependencies in Dependabot.


rahul31124:
Hello @marcnause,

This PR simplifies the code, but the real issue regarding why Dependabot is not creating PRs for Dart (`pub`) dependencies is related to this line in the `pubspec.yaml` file:

https://github.com/fossasia/pslab-app/blob/flutter/pubspec.yaml#L58

This line was added on **September 5, 2025**:

```
dependency_overrides:
  ffi: ^2.1.3
```

The reason for adding it was likely that the **`flusbserial` library depends on `ffi` version `1.x`**, while most modern libraries that use `ffi` require **`ffi 2.x`**. Because of this mismatch, a dependency conflict occurs.

To resolve that conflict, the override was added to force the project to use `ffi 2.x`. However, this override also locks the dependency graph, which prevents **Dependabot** from resolving dependency updates correctly. As a result, Dependabot marks many updates as `update_not_possible` and does not open PRs for `pub` dependencies.

You can see that Dependabot did create a PR earlier on **4th September 2025** and closed it on **5th September 2025**, and it stopped opening dependency update PRs after this override was introduced.
https://github.com/fossasia/pslab-app/pull/2892

It actually took quite a while to track down the real reason 😅, but this appears to be the root cause

Therefore, the possible long-term solutions are:

1. Replace the **`flusbserial`** library with a modern alternative library that supports `ffi 2.x`.
Otherwise, Dependabot will not create PRs for Dart dependency updates.

So while this PR improves the code structure, the root cause of the Dependabot issue lies in the `dependency_overrides` entry for `ffi`.

marcnause:
@rahul31124, I went through the old logs and I think that everything you said is correct. This PR only contains some minor cleanup. Should we keep it anyway or should  I just close it? What do you think?

rahul31124 (changes requested):
@marcnause, the other cleanups are great, but I think using directory: "/" could cause Dependabot to completely ignore our custom composite actions in the .github/actions/* folders. Since Dependabot isn't recursive, please revert the GitHub Actions block back to the plural directories: list

marcnause:
> @marcnause, [...] please revert the GitHub Actions block back to the plural directories: list

Done. I also added several directories which were probably introduced after the dependabot file was edited the last time.

rahul31124 (approved):
LGTM!

CloudyPadmal (approved):

PR #3177, opened 2026-04-13
fix: file saving failure for names exceeding 200 characters

Diksha-MukherjeeAgarwal:
Before:
• If a user inputs a filename longer than 200 characters, the OS rejects it and the file saving process fails. 

After:
• Length Validation & UI Feedback: The filename is validated against a 200-character limit. If the limit is exceeded, a localized Snackbar warning (fileNameTooLong) is displayed to the user immediately, preventing the OS-level crash.

Bug Fixes:
• ​Prevent CSV file saving failures caused by overly long filenames exceeding OS-supported limits..

Enhancements:
​Improve UX by providing immediate, localized UI feedback (Snackbar) when a filename violates length constraints.

## Summary by Sourcery

Enforce filename length limits when saving sensor recordings and provide user feedback for invalid filenames.

Bug Fixes:
- Prevent file saving failures by rejecting filenames longer than 200 characters before passing them to the OS.

Enhancements:
- Add a 200-character limit and trimming to filename inputs across sensor recording dialogs, showing a localized Snackbar warning when the limit is exceeded.
- Align cancel button labeling by removing forced uppercasing in save dialogs.


CloudyPadmal (changes requested):
Rather than truncating, we must **warn** the user and prevent typing longer than 200 characters long names.

lib/view/accelerometer_screen.dart
```diff
@@ -193,11 +194,22 @@ class _AccelerometerScreenState extends State<AccelerometerScreen> {
...
             ),
             ElevatedButton(
               onPressed: () {
-                Navigator.pop(context, filenameController.text);
+                final text = filenameController.text.trim();
+                if (text.length > 200) {
+                  ScaffoldMessenger.of(context).showSnackBar(
+                    const SnackBar(
+                      content: Text(
+                        'File name must be less than 200 characters.',
```
CloudyPadmal: Use localization instead of hardcoded text. That will help out with translation  etc.

CloudyPadmal (changes requested):

lib/view/accelerometer_screen.dart
Diksha-MukherjeeAgarwal: Great catch, I've removed the hardcoded string and added the localization key to the .arb file instead. Applied to all relevant screens.

lib/view/gyroscope_screen.dart
```diff
@@ -184,11 +185,22 @@ class _GyroscopeScreenState extends State<GyroscopeScreen> {
...
             TextButton(
               onPressed: () => Navigator.pop(context),
-              child: Text(appLocalizations.cancel.toUpperCase()),
+              child: Text(appLocalizations.cancel),
             ),
             ElevatedButton(
               onPressed: () {
-                Navigator.pop(context, filenameController.text);
+                final text = filenameController.text.trim();
+                if (text.length > 200) {
```
CloudyPadmal, line 193: As Copilot points out, this check will probably never run as the widget cuts off at set maxlength. So we don't need this clause anymore.

CloudyPadmal (changes requested):
In multiple places

lib/view/accelerometer_screen.dart
```diff
@@ -193,11 +194,22 @@ class _AccelerometerScreenState extends State<AccelerometerScreen> {
...
+                final text = filenameController.text.trim();
+                if (text.length > 200) {
+                  ScaffoldMessenger.of(context).showSnackBar(
+                    SnackBar(
+                      content: Text(
+                        appLocalizations.fileNameTooLong,
+                      ),
+                    ),
+                  );
+                  return;
```
CloudyPadmal, line 210: Could be removed

CloudyPadmal (changes requested):
In two places

marcnause:
Issue fixed in https://github.com/fossasia/pslab-app/pull/3236

PR #3180, opened 2026-04-15
fix: replace filename truncation with user warning

Diksha-MukherjeeAgarwal:
Added validation to prevent users from saving filenames longer than 200 characters and surfacing a warning via Snackbar instead of silently truncating the input.

## Summary by Sourcery

Bug Fixes:
- Prevent filenames longer than 200 characters from being saved by validating input before closing the dialog.


Diksha-MukherjeeAgarwal:
Closing this duplicate PR. All of these changes have been added to the original PR #3177 to keep the commit history and conversation in one place.

PR #3213, opened 2026-04-22
chore: deleted app_localizations files

### LINKED ISSUE #3212: There are many files on GitHub that are generated and updated at build time.
**Describe the issue**

While compiling the project, I noticed that there are many files that are generated and modified with each build. The problem is that they make parsing pull requests and selecting commits more complex.

Questi sono i file che vengono modificati senza che io faccia modifiche quando lancio la build sull'emulatore(per le altre piattaforme ce ne sono altri anche):

Image

Yes, I was thinking of adding as many files as possible to gitignore and deleting them from the repository to make the pulls lighter. (For example, the language localization files.)

ctrlVnt:
Fixes #3212

## Changes
updated .gitignore and deleted app_localization files

If you like the idea, I can apply the exact same thing to epaper and badgemagic.
I'm a bit hesitant to touch all the generated file types, I'd prefer to analyze them slowly first and decide if pushing them is actually pointless

**Checklist**: 


marcnause:
Considering that the app_localitation files are automatically generated at build time and that they cause merge conflicts which need to be resolved manually every now and then, I think removing them from the repo is not a bad idea. I'll think about it a little longer.

marcnause (approved):

PR #3275, opened 2026-05-31
feat(multimeter): add interactive desktop knob controls 

### LINKED ISSUE #3274: Improve desktop usability of the Multimeter knob
Description:
The Multimeter knob is functional on desktop/web, but the interaction could be more intuitive and user-friendly. Improve the desktop experience by adding better interaction feedback and discoverability, such as hover states, cursor changes, tooltips, keyboard accessibility, and other desktop-friendly enhancements where appropriate.

The goal is to make the knob feel more polished and natural to use on desktop platforms while keeping the existing design simple and uncluttered.

> Screenshot for reference 

Image

bhavjsh:
Fixes #3274

## Changes

Improved the Multimeter knob experience on desktop/web.

- Added mouse wheel support to switch between modes.
- Added click support to directly select a mode.
- Kept drag-to-rotate functionality and limited it to the knob area.
- Added keyboard navigation using arrow keys.
- Added accessibility support for screen readers.
- Added smooth pointer animation when changing modes.
- Added grab/grabbing cursor feedback on the knob for desktop.
- Highlighted the currently selected mode for better visibility.

Desktop-specific features are enabled only on desktop/web platforms. Touch devices continue to use tap and drag interactions as before.

## Recordings

https://github.com/user-attachments/assets/065d1baf-6085-41fc-ab43-06ba51949811

### DIFF
--- a/lib/providers/multimeter_state_provider.dart
+++ b/lib/providers/multimeter_state_provider.dart
@@ -116,6 +116,11 @@ class MultimeterStateProvider extends ChangeNotifier {
     notifyListeners();
   }
 
+  void stepMode(int direction) {
+    final count = knobMarker.length;
+    setSelectedIndex((_selectedIndex + direction + count) % count);
+  }
+
   void setSwitch(bool checked) {
     isSwitchChecked = checked;
     if (isSwitchChecked) {
--- a/lib/view/multimeter_screen.dart
+++ b/lib/view/multimeter_screen.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 import 'package:pslab/constants.dart';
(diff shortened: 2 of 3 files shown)


marcnause:
@bhavjsh Nice! :smiley: I love the subtle highlighting of the currently selected mode. :+1: 

My only gripe is that I get a lot of messages like this in the log (running the Linux version of the app):

`Another exception was thrown: A SemanticsNode with action "increase" needs to be annotated with either both "value" and
"increasedValue" or neither`

It would be great if you could take a look.

bhavjsh:
> @bhavjsh Nice! 😃 I love the subtle highlighting of the currently selected mode. 👍
> 
> My only gripe is that I get a lot of messages like this in the log (running the Linux version of the app):
> 
> `Another exception was thrown: A SemanticsNode with action "increase" needs to be annotated with either both "value" and "increasedValue" or neither`
> 
> It would be great if you could take a look.

@marcnause Thanks for the review !  Actually the knob's Semantics was missing increasedValue/decreasedValue, which Flutter needs alongside value. Added them and the exception is gone now :)

marcnause (approved):

PR #3341, opened 2026-06-19
fix: implement missing gyroscope gain

### LINKED ISSUE #3304: Bug: Gyroscope gain is not implemented
**Describe the bug**
GyroscopeConfig defines a sensorGain parameter and exposes it through configuration, but the value is never used in GyroscopeProvider. Changing the gain setting has no effect on gyroscope readings.

**Expected Behavior**
Changing the sensorGain value should scale gyroscope readings before they are displayed. For example, with a gain of 2, a raw reading of 1.5 should be processed as 3.0.

**Steps to reproduce it**
1. Open the Gyroscope experiment.
2. Change the sensor gain setting from its default value (e.g. set it to 5).
3. Start sensor monitoring or recording.
4. Observe the displayed values and recorded data.
5. Compare the output with the output obtained using gain 1.

**Screenshots of the issue**
There is no change between gain 1 and gain 999:

https://github.com/user-attachments/assets/fb4c30ae-b869-4b27-8efb-4f7c45c3f1e1

**System Information**
<table>
<tr>
  <td>Platform</td><td>Android</td>
</tr>
<tr>
  <td>Device</td><td>MOTO G64</td>
</tr>
<tr>
  <td>OS Version</td><td>Android 13</td>
</tr>
</table>

Yes

Jashan32:
Fixes #3304

## Changes
1. Implemented missing implementation of gyroscope gain.
2. Added limit to Gyroscope gain.

## Screenshots / Recordings

https://github.com/user-attachments/assets/bd542942-b87a-4311-8ca7-49df866a7e02

**Checklist**: 


Jashan32:
I've limited the gain range to [0, 100] because the device starts lagging beyond that point. Let me know if we need to support negative gain values as well.

lib/view/gyroscope_config_screen.dart
```diff
@@ -170,9 +170,20 @@ class _GyroscopeConfigScreenState extends State<GyroscopeConfigScreen> {
...
-                        final intValue = int.tryParse(value);
-                        if (intValue != null) {
-                          provider.updateSensorGain(intValue);
+                        final doubleValue = double.tryParse(value);
+                        if (doubleValue != null &&
+                            doubleValue >= 0 &&
+                            doubleValue <= 100) {
+                          provider.updateSensorGain(doubleValue);
+                        } else {
+                          ScaffoldMessenger.of(context).showSnackBar(
```
rahul31124, line 179: @Jashan32, could you move the snackbar to onSubmit instead of onChanged? It might get triggered every time the value changes.

rahul31124 (changes requested):

lib/view/gyroscope_config_screen.dart
Jashan32: @rahul31124 ```ConfigInputItem``` currently calls this callback only when ```OK``` is clicked in the dialog, so it isn't triggered on every keystroke. The snackbar will only be shown on submission.
in ```ConfigInputItem```, ```lib/view/widgets/config_widgets.dart```:
``` dart
TextButton(
  onPressed: () {
    onChanged(controller.text);
    Navigator.of(context).pop();
  },
)
```

rahul31124 (approved):

lib/view/gyroscope_config_screen.dart
rahul31124: Got it, thanks!

marcnause (approved):

PR #3391, opened 2026-07-02
fix: replace unhandled throw with logger.e in FAQ screen launchUrl

### LINKED ISSUE #3384: bug: FAQ screen crashes app when link URL cannot be launched
## Describe the bug
In `lib/view/faq_screen.dart`, `_launchURL()` throws a raw string exception when `canLaunchUrl` returns false:

```dart
Future<void> _launchURL(String url) async {
  final Uri uri = Uri.parse(url);
  if (await canLaunchUrl(uri)) {
    await launchUrl(uri);
  } else {
    throw '${appLocalizations.launchError} $url';
  }
}
```

The only call site is:

```dart
onTap: () => _launchURL(faq.linkUrl!),
```

This is a fire-and-forget call with no try/catch, so the thrown exception is never caught and crashes the app instead of failing gracefully.

## Expected Behaviour
The app should log the failure instead of crashing, consistent with how `about_us_screen.dart` already handles `launchUrl` failures.

## Steps to Reproduce
1. Open the FAQ screen
2. Expand a question that has a link (e.g. "Where to buy")
3. If the link URL cannot be launched, the app crashes

## Possible Fix
Replace the `throw` with `logger.e()`.

## Platform
All platforms

prajakta128:
Fixes #3384

## Problem
`_launchURL()` in `lib/view/faq_screen.dart` throws a raw string exception when `canLaunchUrl` returns false. The only call site is a fire-and-forget `onTap` handler with no `try/catch`, so the exception crashes the app instead of failing gracefully.

## Root Cause
The `_launchURL` method used `throw` for the failure case instead of logging it gracefully. The call site `onTap: () => _launchURL(faq.linkUrl!)` has no try/catch so the exception propagates up and crashes the app.

## Fix
Replaced the `throw` with `logger.e()`, matching the pattern already used in `about_us_screen.dart` for the same kind of failure.

## Impact
- Prevents app crash when a FAQ link cannot be launched
- Consistent with existing logger usage elsewhere in the app
- No UI change

## Screenshots / Recordings
N/A — crash fix, no UI changes

## Checklist

### DIFF
--- a/lib/view/faq_screen.dart
+++ b/lib/view/faq_screen.dart
@@ -39,10 +39,11 @@ class FAQScreen extends StatelessWidget {
         answer: appLocalizations.faqHowToConnectAnswer,
       ),
       FAQItem(
-          question: appLocalizations.faqReportBug,
-          answer: appLocalizations.faqReportBugAnswer,
-          linkText: appLocalizations.faqReportBugLinkText,
-          linkUrl: appLocalizations.faqReportBugLinkUrl),
+        question: appLocalizations.faqReportBug,
+        answer: appLocalizations.faqReportBugAnswer,
+        linkText: appLocalizations.faqReportBugLinkText,
+        linkUrl: appLocalizations.faqReportBugLinkUrl,
+      ),
       FAQItem(
         question: appLocalizations.faqRecordData,
         answer: appLocalizations.faqRecordDataAnswer,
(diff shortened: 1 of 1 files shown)


rahul31124 (commented):
@prajakta128 Which issue did you solve here?

prajakta128:
Hi @rahul31124, this PR fixes issue #3384 — the FAQ screen was crashing when a link URL could not be launched because `_launchURL()` was using `throw` instead of handling the error gracefully.

However, I just noticed that this same fix was already merged in my earlier PR #3385. This PR #3391 is a duplicate. I am closing it now.

Sorry for the confusion!

PR #3393, opened 2026-07-02
fix: replace hardcoded 'Not Connected' string with localized value

### LINKED ISSUE #3392: bug: hardcoded 'Not Connected' string in BoardStateProvider ignores app locale
## Describe the bug
In `lib/providers/board_state_provider.dart`, the string `'Not Connected'` is hardcoded in three places:

- Line 20: `String pslabVersionID = 'Not Connected';`
- Line 63: `pslabVersionID = 'Not Connected';`
- Line 77: `pslabVersionID = 'Not Connected';`

This means the board connection status always shows in English regardless of the user's chosen language in the app. All other languages are ignored for this string.

## Expected Behaviour
The string should use `appLocalizations.notConnected` (or equivalent localization key) so it respects the user's chosen language.

## Steps to Reproduce
1. Set the app language to any non-English language
2. Connect and disconnect a PSLab device
3. The status shows "Not Connected" in English instead of the translated string

## Platform
All platforms

prajakta128:
Fixes #3392

## Problem
`pslabVersionID` in `BoardStateProvider` was hardcoded to the English string `'Not Connected'` in three places:

- Field declaration (line 20)
- USB detach event handler
- WiFi disconnect handler

This means all non-English users always see "Not Connected" in English regardless of their chosen app language, breaking the app's internationalization (i18n) support.

## Root Cause
The string was hardcoded directly instead of using the existing localization key. The `notConnected` key already exists in `lib/l10n/app_en.arb` and all other supported language ARB files, so no new translations were needed.

## Fix
Replaced all three instances of `'Not Connected'` with `appLocalizations.notConnected`. Also changed the field declaration default from `'Not Connected'` to `''` since `appLocalizations` is a getter and cannot be used at field declaration time.

## Impact
- Non-English users now see the correct translated string for the disconnected state
- No functional change for English users
- No new translation keys needed — `notConnected` already exists in all language files

## Screenshots / Recordings
N/A — localization fix, no visual change for English users

## Checklist

## Summary by Sourcery

Localize the board disconnection status text and tidy related provider logic.

Bug Fixes:
- Use the localized `notConnected` string for the PSLab disconnection status so it respects the app language settings.

Enhancements:
- Initialize the PSLab version ID field with an empty string and slightly refactor event listener formatting for USB and connectivity handlers.

### DIFF
--- a/lib/providers/board_state_provider.dart
+++ b/lib/providers/board_state_provider.dart
@@ -29,6 +29,7 @@ class BoardStateProvider extends ChangeNotifier {
   BoardStateProvider() {
     scienceLabCommon = getIt.get<ScienceLabCommon>();
     configProvider = SettingsConfigProvider();
+    pslabVersionID = appLocalizations.notConnected;
   }
 
   Future<void> initialize() async {
--- a/lib/providers/board_state_provider.dart
+++ b/lib/providers/board_state_provider.dart
@@ -17,7 +17,7 @@ class BoardStateProvider extends ChangeNotifier {
   bool pslabIsConnected = false;
   bool hasPermission = false;
   late ScienceLabCommon scienceLabCommon;
-  String pslabVersionID = 'Not Connected';
+  String pslabVersionID = '';
   String pslabVersionIDV6 = 'PSLab V6';
   String pslabVersionIDV5 = 'PSLab V5';
   int pslabVersion = 0;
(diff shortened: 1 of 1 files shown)


rahul31124 (commented):
Hello @prajakta128, these strings are intended to remain as they are because they are logical strings, not UI strings. Therefore, these changes are not relevant. I'm closing this issue. Thanks for the contribution!

